PR #1, opened 2017-10-19
[#151838550] Create application

bandesz:
## What

This PR adds a Golang application for polling the CloudFoundry API for app and service usage events and storing them in Postgres.

Cloud Foundry docs on usage events: https://docs.cloudfoundry.org/running/managing-cf/usage-events.html

The application periodically fetches the latest app and service usage events from the following endpoints:
 * /v2/app_usage_events - (API docs)[http://apidocs.cloudfoundry.org/272/app_usage_events/list_all_app_usage_events.html]
 * /v2/service_usage_events - (API docs)[https://apidocs.cloudfoundry.org/272/service_usage_events/list_service_usage_events.html]

The API endpoints are paginated and we are using the *after_guid* parameter to get the next batch of events. We are not processing the latest events (COLLECTOR_RECORD_MIN_AGE) as these can be incomplete because events are only present if the related transactions are finished. (The events are ordered as they are started.)

Note: The API lists the events in order but the timestamps on events should not be used to sequence events (local clock issue in a multi-server environment).

The application will be deployed as a Cloud Foundry application and ships with a manifest.yml file.

The PaaS PR can be found [here](https://github.com/alphagov/paas-cf/pull/1073)

## How to review

 * consult the README and try to run the application locally
   * the app logs should contain multiple json lines with record_count > 0
   * he postgres database should contain records in the app_usage_events and service_usage_events tables
   * the usage events should be inserted regularly expect for events newer than 5 minutes
 * check if the README contains all necessary information

## Who can review

not me or @henrytk


manifest.yml
```diff
@@ -0,0 +1,12 @@
+---
+applications:
+ - name: paas-usage-events-collector
+   memory: 128M
+   disk_quota: 100M
+   instances: 1
+   buildpack: go_buildpack
+   no-route: true
```
samcrang: I think this is causing deploys in my dev environment to fail. I get the following error:

```
Could not unmap route - Route paas-usage-events-collector-new not found
```

It looks as though `blue-green-deploy` relies on having a predictable route based off the application name?

alext: For an app without a route, there shouldn't be a need to do a blue-green deploy at all should there?

bandesz: You are right, I changed it to cf push instead.

README.md
```diff
@@ -1,2 +1,90 @@
...
+
+### Regenerating the mocks
+
+The tests are using generated mocks created with [gomock](https://github.com/golang/mock). If you change any of the interfaces you have to regenerate the mocks with the provided **generate-test-mocks** make target:
+
+```
+make generate-test-mocks
+```
+
+## Run application locally
```
samcrang: I'd maybe consider bumping this section up one or mentioning that you need to have a postgres instance available in order to run the tests with `make test`.

PR #5, opened 2017-11-24
[#152241494] Billing API

chrisfarms:
## What

We are extending the event collector application to give us a "Billing API" that allows us to query resource usage and calculate costs for that usage. It may make sense to rename this repository "paas-billing" in the future, and use it to contain those items that are dependent on the billing database.

* A `MATERIALIZED VIEW` is created in the database that gets updated at a set interval (currently 1hr) which normalizes the raw event data into rows of `resource_guid`, `time period`, `pricing_plan_id` , then joins the rows with data from a `pricing_plans` table that contains the information required to calculate the prices. The view caches and indexes data in the following form so that we can perform queries against it efficiently. 

| ... | duration | guid | org | space | pricing_plan_id | price | ... |
| --- | --- | --- | --- | --- | --- | --- | --- |
| ... | `2017-11-01 08:14` to `2017-11-19 08:14` | {thing1} | {org1} | {space1} | {planA} | £x | ... |
| ... | `2017-11-01 09:10` to `2017-11-19 10:10` | {thing2} | {org2} | {space1} | {planB} | £x | ... |

* A `pricing_plans` table is added to the database that contains the formulas required to calculate costs for each resource. For example a pricing_plan row might look like:

| id | name | valid_from | plan_guid | formula |
| --- | --- | --- | --- | --- |
| 2 | compute instance | 2017-11-01 | {plan_guid} | `($memory_in_mb / 1000) * ($time_in_seconds)` |
| 3 | tiny postgres | 2017-11-01 | {plan_guid} | `1 * $time_in_seconds` |

* Pricing plans can change over time so they have a valid_from field. The view calculation handles splitting usage over the valid ranges.

* A REST/JSON API exposes aggregated data at several levels. Only guid details are returned in the data at the moment. If you want names you would need to call out the cf:
    - `/organisations` list totals for all orgs
    - `/organisations/:org_guid` list total for a single org
    - `/organisations/:org_guid/spaces` list totals for each space in an org
    - `/organisations/:org_guid/resources` list totals for each resource in org (a resource is a single thing - like an "app" or a "service")
    - `/spaces` list totals for all spaces
    - `/spaces/:space_guid` list total for a single space
    - `/spaces/:space_guid/resources` list totals for each resource in space
    - `/resources` list totals for all resources
    - `/resources/:resource_guid` list total for a single resource
    - `/events` [experimental] list all events ("events" are all the start/stop points with calculated billing, unlike "resources" which are aggregate totals for each item over a range). Events would allow you to see _when_ something happens
    - `/resources/:resource_guid/events` as above but for a single resource
    - `/pricing_plans` fetch the pricing plans

* A (throwaway) example HTML rendering of an aggregated report can be found at `/` you will be prompted to login via UAA. This is meant purely as an illustration of what is possible for now.

## How to review

* Deploy to dev using https://github.com/alphagov/paas-cf/pull/1116
* Go start/stop some apps/services to generate some events in your dev environment (ie. restart the healthcheck apps, create a postgres service etc
* Visit https://paas-usage-events-collector.$DEPLOY_ENV.dev.cloudpipelineapps.digital/
* Login as an admin when prompted and authorize the app 
* Visit /pricing_plans and create a pricing plan for the compute plan with guid: `f4d4b95a-f55e-4593-8d54-3364c25798c4`
* Visit /pricing_plans and press the "Seed pricing plans" button to create initial pricing plans for all known services.
* Visit `/` to see an HTML report of some pricing
* Query the Billing API for some results

## Who can review

Not @chrisfarms, @46bit, @camelpunch, @samcrang


LeePorte:
Thoughts

 - Manifest has `no-route :true`
 - Starting with an empty db causes the app to crash
 - Lacking in security, it uses basic auth at present. I have not checked for other issues as yet
 - The app currently doesn't bind to the port provided by CF, this has restricted my testing to using the local method so far.
 
App and service events have been recorded and I have been able to run queries against them whilst working locally.

.editorconfig
```diff
@@ -0,0 +1,18 @@
+# Use editorconfig to stop whitespace wars
```
henrytk: > Use editorconfig to stop whitespace wars

Sounds like loser talk.

API.md
```diff
@@ -0,0 +1,33 @@
+## API usage examples
+
```
henrytk: This file is really useful. Two points:

* there are also lots of useful bits in the pull request description that could go in here. I don't want to have to go find the pull request every time.
* this file isn't very discoverable. Could it be part of the README, or at least linked to by the README?

db/repair.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+// RepairEvents populates START events for applications and services that are
+// missing due to the events being emitted before data collection began (T0).
+//
+// * If an application guid is currently in a running state according to
+//   CloudController, but no events exist, then a START event is created for T0
+// * If a service_instance_guid is currently in a CREATED state according to
+//   CloudController, but no events exist, then a CREATED event is created for T0
+func (pc *PostgresClient) RepairEvents(cfClient cf.Client) (err error) {
```
henrytk: I can't find where this actually gets called from. Where is it used? `ack 'RepairEvents' -i` turned up nothing in the entire repository (other than lines 12 and 19 of this file of course).

db/sql/030_create_billable.sql
```diff
@@ -0,0 +1,124 @@
...
+		select
+			id,
+			created_at::timestamptz as created_at,
+			(raw_message->>'app_guid') as guid,
+			(raw_message->>'app_name') as name,
+			(raw_message->>'org_guid') as org_guid,
+			(raw_message->>'space_guid') as space_guid,
+			'f4d4b95a-f55e-4593-8d54-3364c25798c4'::text as plan_guid, -- fake plan id for compute plans
+			'default-compute'::text as plan_name,                      -- fake plan name for compute plans
+			coalesce(raw_message->>'instance_count', '0')::numeric as inst_count,
```
henrytk: When do we expect this to be null? Same goes for `memory_in_mb_per_instance` on the line below.

db/sql/030_create_billable.sql
```diff
@@ -0,0 +1,124 @@
+DROP MATERIALIZED VIEW IF EXISTS billable;
+CREATE MATERIALIZED VIEW IF NOT EXISTS billable AS with
+
+-- extract useful stuff from usage events
+-- we treat both apps and services as billable "resources" so normalize the fields
+-- we normalize states to just STARTED/STOPPED because we treat consecutive STARTED to mean "update"
+events as (
```
henrytk: `events` doesn't seem to select `previous_instance_count` or `previous_memory_in_mb_per_instance`, so how does it produce data we can calculate scaling events from?

fixtures/plans.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/alphagov/paas-usage-events-collector/db"
+)
+
+type Plans map[string]Plan
+
+type Plan struct {
+	ID        int
+	PlanGuid  string
+	ValidFrom time.Time
+	Formula   string
```
henrytk: These formulas are really cool.

henrytk: These formulas are a lot to get your head around.

API.md
chrisfarms: 🙌  on both parts.

I'll improve the content and merge into README.

I guess I was unsure what level of "done" we are going for :)

db/repair.go
chrisfarms: Good point! 

It was originally executed on startup ... but that was a bit presumptuous. It should be exposed via the API or removed entirely (it's only required as an alternative solution to triggering the cloud controller "pure and reset events" thing).

I'll had a handler for it and document it, then it stays as an option for now (since I found it useful for fixing  a dump of events from prod).

db/sql/030_create_billable.sql
chrisfarms: It's not expected to be null ... but there's nothing currently _stopping_ it from being null and we want that column to always be not null. Bit overprotective maybe.

`(select (NULL + 1) is null)` => `null`  ...which would be a bad thing for calcuations.

We could get the same behaviour by adding an index/trigger to validate the incoming events data, but decided that handling the error cases of a bad incoming event was much harder than filtering the data.

db/sql/030_create_billable.sql
chrisfarms: Everything is operating on the idea that we have "good" data.

We can make that assumption because we are going generate all the start events using either the `RepairEvents` method or by call the "purge and reset events" cloud controller equivalent.

So we are not "guessing" previous states. Usage is calculated from a `START` event to the next `START` or `STOP` event.

db/repair.go
chrisfarms: Creating a decent `RepairEvents` function that makes the best use of the data we have to create the most accurate fake events it can is a story in itself.

.editorconfig
chrisfarms: https://www.youtube.com/watch?v=SsoOG6ZeyUI&feature=youtu.be&t=62s

db/sql/030_create_billable.sql
henrytk: So, if in the last 5 minutes of the calendar month I scale my up then down again I only get billed for 5 minutes?

chrisfarms: No. 

I think maybe you are framing the problem in terms of "months", and this is causing some confusion.

We have access to ALL of the events from the moment we started collecting them, and we are working under the assumption that all this data is "good".

So in the scenario you describe, we may not have seen a `START` event during _this_ month, but we will have seen one before that. So when a query is made that requests the billable resources limited to a slice a time, ie between t1 and t2 it will take that into account.

Does that make sense?

server/server.go
```diff
@@ -0,0 +1,86 @@
...
+	e.POST("/pricing_plans", auth.AdminOnly(api.CreatePricingPlan(db)))
+	e.PUT("/pricing_plans/:pricing_plan_id", auth.AdminOnly(api.UpdatePricingPlan(db)))
+	e.DELETE("/pricing_plans/:pricing_plan_id", auth.AdminOnly(api.DestroyPricingPlan(db)))
+	e.POST("/seed_pricing_plans", auth.AdminOnly(api.CreateMissingPricingPlans(db)))
+
+	return e
+}
+
+func redirectToReport(c echo.Context) error {
+	lastMonth := time.Now().UTC().Add(-30 * 24 * time.Hour).Format(time.RFC3339)
```
henrytk: This could be a bit more user friendly by selecting the first day of the month:

```
now := time.Now().UTC()
lastMonth := time.Date(now.Year(), now.Month()-1, 1, 0, 0, 0, 0, now.Location()).Format(time.RFC3339)
```

henrytk: Unless a rolling 30-day thing was what you were after

db/sql/030_create_billable.sql
henrytk: ah, awesome, thanks.

server/server.go
chrisfarms: I was going for a "show me recent usage to get started" type affair, so there was some example date ranges to play with.

The whole "report" and html ui thing is a bit of a sketch tbh ...it's more of a "you could make a UI that does things like this" than a "this is a ui we are going to keep". It was added because it was tiresome having to make API calls directly to play with the output and because without it I didn't really see a way for Tom to poke around.

I suspect all this functionality to end up in the propper "paas admin dashboard/control-panel" ruby app thing eventully.

bandesz:
It seems to me we run the database migration queries at every start which might be not optimal, because we drop the billing view every time before recreating it. Probably we should store which migrations ran in the db so we run only the new ones.

chrisfarms:
Yeah it's not really a "migration" at the moment... more like a set of definitions executed in order.

I agree we could improve it and have a table for storing where we are in the migration, however since I expect we will be tweaking these things a fair bit in coming iterations, and since most of the things in the "migration" can safely be destroyed and recreated without issue (functions/views), this was preferable than ending up with lots of migrations that redefine the same things over and over with slight changes.

The main "billable" view gets recreated every hour anyway, and doing it on startup (to get the latest version on first start) would be less confussing than altering the view but getting the old one for an hour on start.

server/server.go
henrytk: Cool. Leave as is. We can do code tweaks when people start actually using the thing.

api/api_test.go
```diff
@@ -0,0 +1,1004 @@
...
+							"app_guid": "app2",
+							"app_name": "app_name2",
+							"org_guid": "org_guid2",
+							"space_guid": "space_guid2",
+							"instance_count": 1,
+							"memory_in_mb_per_instance": 512,
+							"previous_state": "STOPPED"
+						}`),
+					}, {
+						MetaData: cf.MetaData{CreatedAt: now.Add(-61 * time.Minute)},
```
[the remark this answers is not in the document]
chrisfarms: The idea here was to NOT test exactly the boundary :)

db/repair.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return err
+	}
+
+	// fetch the most recent event timestamp
+	var firstEventTimestamp *time.Time
+	err = tx.QueryRow(`
+		with
+		event_dates as (
+			select created_at from app_usage_events
```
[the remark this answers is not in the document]
chrisfarms: Yeah

```
select least((select min(created_at::timestamptz) from app_usage_events where created_at is not null), (select min(created_at::timestamptz) from service_usage_events where created_at is not null));
```

Would likly be 2 or 3 times faster.

db/repair.go
```diff
@@ -0,0 +1,264 @@
...
+		if err != nil {
+			return err
+		}
+		fmt.Println("app", app.Guid)
+		// Skip if we already have events for this app
+		if seen := knownGuids[app.Guid]; seen {
+			continue
+		}
+		fmt.Println("detected missing app", app.Guid)
+		if updated.After(*firstEventTimestamp) {
```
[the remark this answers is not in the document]
chrisfarms: yes we _should_ have an event, so we skip this app since if for some reason we got to this point then starting creating fake events it would be incorrect.

I'm not sure exactly when that case _could_ occur... I only know that it should be skipped if it _did_ occur :)

One possible situation is if an app was updated but the event hasn't arrived yet.

db/repair.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return err
+	}
+	for _, serviceInstance := range srvs {
+		// Skip anything updated since we started collecting events
+		updated, err := time.Parse(time.RFC3339, serviceInstance.UpdatedAt)
+		if err != nil {
+			return err
+		}
+		if updated.After(*firstEventTimestamp) {
```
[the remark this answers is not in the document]
chrisfarms: These are all excellent questions. And are one of the reasons this function isn't called heh.

We're going to have two choices:

1. press the "purge and reset" button and consider data "good" from that point in time
2. OR implement something like RepairEvents that takes into account the data we _do_ have and makes it as accurate as possible.

API.md
```diff
@@ -0,0 +1,33 @@
...
+### Authorization
+
+All API request require a valid token set in the `Authorization` header.
+
+To get a valid token you must sent a requst to `/oauth/callback?code=[AUTH_CODE]&state=[AUTH_STATE]`
```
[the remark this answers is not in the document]
chrisfarms: I on't hav thos kys on my keyboar

db/sql/020_create_service_usage_events.sql
```diff
@@ -5,3 +5,5 @@ CREATE TABLE IF NOT EXISTS service_usage_events (
 	raw_message JSONB
 );
 CREATE INDEX IF NOT EXISTS idx_service_usage_id ON service_usage_events (id);
+CREATE INDEX IF NOT EXISTS idx_service_usage_state ON app_usage_events ( (raw_message->>'state') );
+CREATE INDEX IF NOT EXISTS idx_service_usage_type ON app_usage_events ( (raw_message->>'service_instance_type') );
```
bandesz: both indexes should be created on the service_usage_events table

server/server.go
```diff
@@ -59,6 +60,12 @@ func New(db db.SQLClient, authority auth.Authenticator) *echo.Echo {
 	e.DELETE("/pricing_plans/:pricing_plan_id", auth.AdminOnly(api.DestroyPricingPlan(db)))
 	e.POST("/seed_pricing_plans", auth.AdminOnly(api.CreateMissingPricingPlans(db)))
 
+	// Repair events
+	if cf != nil {
+		e.GET("/repair_events", auth.AdminOnly(api.GetRepairedEvents(db, cf)))
+		e.POST("/repair_events", auth.AdminOnly(api.RepairEvents(db, cf)))
```
bandesz: when I call this with curl (the POST endpoint) I get into an infinite redirect loop to /repair_events

bandesz:
Two remaining issues:
 * new indexes were created on the wrong table
 * if I post to repair events it results in an infinite redirect loop and doesn't do anything (tried with curl)

chrisfarms:
Good catch on indexes ... I didn't notice in the query plan output as I assumed it was doing a seq scan due to the small size of the service events table!... fixed

I believe the redirect loop is due to curl. Did you use `-X POST` by any chance? ... since I think that forces it to follow redirects using POST too, which is weird and would cause the behaviour you describe.... you can use `-d ""` instead and it should do as you expect.

Wether or not it "does anything" is somewhat dependant on the state of your data. You would probably have to manually delete events from the database so that there are some missing.

db/sql/020_create_service_usage_events.sql
chrisfarms: Good catch.

server/server.go
chrisfarms: I blame `curl -X POST` behaviour. Try using `-d ""` instead.

bandesz:
@chrisfarms yeah, you were right with curl - I was expecting something like that. Thanks for the other fix.

Notes for repair for the follow-up story:
 - we should insert missing start events for apps where the first seen event was a stopped one (or even an update one), the current mechanism won't do this
 - we should insert missing start events for services where the first seen event was a stopped one (we can't do anything about an update as we don't know the previous state)

Great work, I will approve this now.

@chrisfarms could you please squash the commits before I merge this?

PR #6, opened 2017-12-13
[#153508326] Repair events tool

chrisfarms:
## What

We have built a cli tool `db_repair` that can be used to generate missing STARTED/CREATED events so that we can trust our queries against the resource usage data.

See the [db_repair README](https://github.com/alphagov/paas-usage-events-collector/tree/db_repair_cmd/cmd/db_repair) and commit messages for more details.

We hope that we will not need to run this tool often (if ever) but decided to make it a tool that can be run multiple times in case we find more edge cases in the future.

## How to review 

The reviewer should ensure that the repair process makes sense and will do what it says on the tin.

We have already used the tool to "repair" the production dataset. But the process is reversible and so we can modify and re-run the tool if any problems are found.

## Who can review

Not @dcarley @henrytk @chrisfarms


cmd/db_repair/repair.go
```diff
@@ -232,25 +212,174 @@ func (pc *PostgresClient) RepairEvents(cfClient cf.Client) (err error) {
...
+	}
+	if firstEventTimestamp == nil {
+		return nil, errors.New("Database appears to be empty and thus cannot be repaired.")
+	}
+	return firstEventTimestamp, nil
+}
+
+func createEventsForAppsWhereFirstRecordedEventIsStopped(tx *sql.Tx, firstEventTimestamp *time.Time) (err error) {
+	result, err := tx.Exec(`
+		WITH events AS (
```
bandesz: Not important: this seems quite expensive to me as for one application there will be multiple lines. Is there a way to use group by here for the app_guids and simply select the first raw_message (ordered by id)

cmd/db_repair/README.md
```diff
@@ -0,0 +1,68 @@
...
+* No recorded event for a service we know is running in cf
+* First recorded event for an is STOPPED
+* First recorded event for a service is DELETED
+
+The following cases are intensionally ignored:
+
+* If we saw a service UPDATED event but without a preceeding CREATED: This is
+ignored as we cannot predict the previous state data
+* If we saw an app update event (STARTED with previous state STARTED):
+This is ignore as we do not consider the impact to be too servere.
```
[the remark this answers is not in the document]
henrytk: The kick-off notes reflect the missing data we agreed were important enough to focus on. We acknowledged that there were many more edge cases you could spend time on, but it unclear whether the cost of developer time to fix everything is worth it.

cmd/db_repair/repair.go
henrytk: Could you elaborate on 'simply select the first raw_message' please? What would be doing the selecting? It would be good if there was a less expensive way of doing this.

bandesz: I don't know Postgres as well. After a quick google it seems we could use the row_number() function on the window and include a condition in where to include only the lines with row_number = 1

bandesz: https://stackoverflow.com/a/33391656/873178

henrytk:
@bandesz 

> From the ticket: 183 vs 184 apps discrepancy - where is that extra one app coming from when you ran the script? (the query returned with 183 apps without events but the script inserted 184)

It skips non-running apps, so one possible cause is an app wasn't running when we were testing, but was running when we ran it for real.

I think what we're really missing here is a method for integrity checking the database. In hindsight we should have devised ways of producing stats on broken data before trying to fill the gaps.

henrytk:
> When creating the missing apps we could set the buildpack data

Good spot. I've added the fields.

henrytk:
> Just a note: we have to make sure our bills are consistent so I would be careful rerunning the script with new edge cases in the future.

Agreed. If we discover a need to use this again we should be considering various ways of making it more reliable: integrity checks, tests of the code _et cetera_.

henrytk:
> Why should we skip services and apps where we have no events but the updated timestamp is after we started collecting data?

I have forgotten why we are doing this. Any clue @chrisfarms ?

chrisfarms:
> Why should we skip services and apps where we have no events but the updated timestamp is after we started collecting data?

Since this implies that the event collector has simply not collected the event yet due to the collection delay.

Admittedly the time range for this check is a bit wider than needed, but I believe this does not affect the logic.

cmd/db_repair/repair.go
chrisfarms: You could get the same effect by just combining the `events` and `stop_events` CTEs into a single query.... but we had split them out this way for readability over performance since this is a very infrequently run task :)

cmd/db_repair/repair.go
```diff
@@ -85,46 +38,28 @@ func (pc *PostgresClient) RepairEvents(cfClient cf.Client) (err error) {
...
 		if err != nil {
 			return err
 		}
-		fmt.Println("app", app.Guid)
-		// Skip if we already have events for this app
-		if seen := knownGuids[app.Guid]; seen {
-			continue
-		}
-		fmt.Println("detected missing app", app.Guid)
 		if updated.After(*firstEventTimestamp) {
```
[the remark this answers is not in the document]
chrisfarms: This is about not creating fake events for apps that have already or are about to emit STARTED events on their own.

It's possible that checking the "created at" would be more appropriate, but updated seemed safer to me, since creating an bad fake event (and charging more) is worse than skipping the event (and undercharging).

bandesz:
Could you please squash the commits? After that feel free the merge the PR.

PR #7, opened 2018-01-17
[#153435158] Create a statement/report endpoint for the Billing API

LeePorte:
## What

We have added a reporting / billing endpoint that is filtered by Organisation GUID, this simplifies the bill generation process significantly and can now be performed by the Product team using any set of dates.

## How to review

**This requires production access to test**

### Dump production billing data

```
./scripts/cf_subshell_scoped_login.sh prod
cf target -o admin -s billing
cf conduit billing-db -- \
  docker run -ti \
    -e PGDATABASE \
    -e PGUSER \
    -e PGPASSWORD \
    -e PGPORT \
    -e PGHOST=docker.for.mac.localhost \
    --name billing-dump \
    postgres:9.6-alpine \
    pg_dump -Fc -f billing.dump
docker cp billing-dump:billing.dump .
docker rm -f billing-dump
```

### Start the database server

```
docker run --name billing -e POSTGRES_PASSWORD= -p 5432:5432 -d postgres:9.6-alpine
```

### Import the billing data dump
```
docker cp billing.dump billing:/
docker exec -ti billing pg_restore --create --clean --no-privileges --no-owner -f billing.sql billing.dump
vi billing.sql # remove reassign_owned() references
docker exec -ti billing sh -c 'psql -U postgres < billing.sql'
```

### Change client in your dev environment
```
cd scripts && bundle exec uaac client update paas-usage-events-collector --redirect_uri http://localhost:8881/oauth/callback
```

### Run the billing app locally
```
DATABASE_URL=postgres://postgres:@localhost:5432/rdsbroker_b946a69b_76f4_4cbc_80df_5cee2d54b7ad?sslmode=disable make run-dev
```

### Local testing
Visit the `/report/55b1eb7d-e4c5-4359-9466-dd3ca5b0e457` endpoint
With a date range set to 2017-01-01T00:00:00Z to 2018-01-01T00:00:00Z you should see an org total of £5504.69

## Who can review

Anyone with production access, but not @LeePort or @dcarley

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -44,7 +44,7 @@ func NewReportHandler(db db.SQLClient) echo.HandlerFunc {
 			with
 			resources as (
 				select
-					guid,
+					name,
 					space_guid,
 					pricing_plan_id,
 					pricing_plan_name,
@@ -55,9 +55,9 @@ func NewReportHandler(db db.SQLClient) echo.HandlerFunc {
 				where
 					org_guid = $1
 				group by
-					guid, space_guid, pricing_plan_id, pricing_plan_name
+					name, space_guid, pricing_plan_id, pricing_plan_name
 				order by
-					guid, space_guid, pricing_plan_id
+					name, space_guid, pricing_plan_id
 			),
 			space_resources as (
 				select
--- a/api/api.go
+++ b/api/api.go
@@ -36,12 +36,15 @@ func NewUsageHandler(db db.SQLClient) echo.HandlerFunc {
 
 func NewReportHandler(db db.SQLClient) echo.HandlerFunc {
 	return func(c echo.Context) error {
-		return withAuthorizedResources(Many, c, db, `
+		orgGUID := c.Param("org_guid")
+		if orgGUID == "" {
+			return errors.New("missing org_guid")
+		}
+		return withAuthorizedResources(Single, c, db, `
 			with
 			resources as (
 				select
 					guid,
-					org_guid,
 					space_guid,
 					pricing_plan_id,
 					pricing_plan_name,
@@ -49,35 +52,32 @@ func NewReportHandler(db db.SQLClient) echo.HandlerFunc {
 					sum(price * 100)::bigint as price
 				from
 					authorized_resources
+				where
+					org_guid = $1
 				group by
-					guid, space_guid, org_guid, pricing_plan_id, pricing_plan_name
+					guid, space_guid, pricing_plan_id, pricing_plan_name
 				order by
-					guid, space_guid, org_guid, pricing_plan_id
+					guid, space_guid, pricing_plan_id
 			),
 			space_resources as (
 				select
-					t.org_guid,
 					t.space_guid,
 					sum(t.price) as price,
 					json_agg(row_to_json(t.*)) as resources
 				from
 					resources t
 				group by
-					org_guid, space_guid
+					space_guid
 				order by
-					org_guid, space_guid
+					space_guid
 			)
 			select
-				t.org_guid,
+				$1 org_guid,
 				sum(t.price) as price,
 				json_agg(row_to_json(t.*)) as spaces
 			from
 				space_resources t
-			group by
-				org_guid
-			order by
-				org_guid
-		`)
+		`, orgGUID)
 	}
 }
 
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -848,7 +848,31 @@ var _ = Describe("API", func() {
 					EntityRaw: json.RawMessage(`{
 						"state": "STOPPED",
 						"app_guid": "o1s1-app1",
+						"app_name": "o1s1-app1-renamed",
+						"org_guid": "o1",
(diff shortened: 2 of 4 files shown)


keymon:
Code LGTM. I cannot test it as I do not have access to prod, but I am going to test it locally.

I have some doubts that I would walk through with any of you:
 - Is it guarantee that the `name` of the `billable` view would always match the first event?
 - outside of the scope of this PR, I have doubts about when the views get refreshed and if we should change that.

keymon:
Travis build is broken, I do not have permissions to retrigger atm. it likely would pass, but if anyone can do it?

dcarley:
> Code LGTM. I cannot test it as I do not have access to prod, but I am going to test it locally.

I've sent you a dump of the prod data. It's already been filtered for roles and triggers, so you don't need to run the `pg_restore` step.

> Travis build is broken, I do not have permissions to retrigger atm. it likely would pass, but if anyone can do it?

I've re-triggered.

> Is it guarantee that the name of the billable view would always match the first event?

Not sure I follow. Can you explain further?

> outside of the scope of this PR, I have doubts about when the views get refreshed and if we should change that.

It gets updated once an hour. Does this help?

- https://github.com/alphagov/paas-usage-events-collector/blob/2ba404b5dc64af3987706dff079d859756b5db40/main.go#L98-L113
- https://godoc.org/github.com/alphagov/paas-usage-events-collector/db#PostgresClient.UpdateViews

keymon:
>> Is it guarantee that the name of the billable view would always match the first event?
> Not sure I follow. Can you explain further?

In the commit 0292302275bf7e9c9e5f3dbda1c41ab7747b5ed2 we group and return the application name rather than the guid, in order to use always the name of the application on the first "STARTED" event. 

But it was not so clear for me that the `name` attribute would be the one from the `STARTED` event.  

Anyway, after talking with @chrisfarms the name comes from each event when creating the `event_ranges`: https://github.com/alphagov/paas-usage-events-collector/blob/ae26208342f34e1cba9c4615cefbefa4b5238d71/db/sql/030_create_billable.sql#L52-L72, and then we only take into account the `STARTED` events in https://github.com/alphagov/paas-usage-events-collector/blob/ae26208342f34e1cba9c4615cefbefa4b5238d71/db/sql/030_create_billable.sql#L114-L116

dcarley:
> But it was not so clear for me that the name attribute would be the one from the STARTED event.

Do you think we could improve the tests to make this clearer?

keymon:
I noticed that if you visit a report without authentication: http://localhost:8881/report/123456, the user would be sent to authenticate to UAA, but once authenticated, it gets back to http://localhost:8881/report WITHOUT the org ID.

That is because the logic in https://github.com/alphagov/paas-usage-events-collector/blob/ae26208342f34e1cba9c4615cefbefa4b5238d71/server/server.go#L39-L40 and
https://github.com/alphagov/paas-usage-events-collector/blob/ae26208342f34e1cba9c4615cefbefa4b5238d71/server/server.go#L65-L69 , which I think it does not make sense anymore.

We should remove that redirect, or, otherwise, add logic in the [`UAATokenAuthentication()`](https://github.com/alphagov/paas-usage-events-collector/blob/ae26208342f34e1cba9c4615cefbefa4b5238d71/auth/handler.go#L20-L21) to perserve the original URL.

keymon:
Other issue, the UI does not work when rendering the report, with the following error:
```
{"time":"2018-01-18T12:37:34.584518091Z","level":"ERROR","prefix":"echo","file":"echo.go","line":"284","message":"template: base:59:49: executing \"content\" at <in_pounds>: wrong number of args for in_pounds: want 1 got 0"}
```

keymon:
I am happy with the PR, apart of the issues with the UI (see comments above):
 * Auth redirects to `/report` which does not work anymore.
 * the "update range" in UI sends to `/report` which does not work anymore
 * The report does not render HTML with the error `<in_pounds>: wrong number of args for in_pounds: want 1 got 0`

dcarley:
> Auth redirects to /report which does not work anymore.
> the "update range" in UI sends to /report which does not work anymore
Thanks, I'll take a look at those.

> The report does not render HTML with the error <in_pounds>: wrong number of args for in_pounds: want 1 got 0
We didn't notice this in our testing. Is it producing any HTML output? Can you describe how to reproduce?

keymon:
>>  The report does not render HTML with the error <in_pounds>: wrong number of args for in_pounds: want 1 got 0

> We didn't notice this in our testing. Is it producing any HTML output? Can you describe how to reproduce?

It produces some HTML, but appears empty. Simply go to a report url with a browser: http://localhost:8881/report/55b1eb7d-e4c5-4359-9466-dd3ca5b0e457

dcarley:
I've addressed all of the comments with some additional PRs.

We couldn't reproduce the `in_pounds` error for valid organisations, but I've guarded against it when the org isn't found.

keymon:
LGTM :+1:

PR #11, opened 2018-02-15
 [#154329640] add pricing estimation endpoint

keymon:
[#154329640 Pricing Calculator: Implement price estimation via billing API](https://www.pivotaltracker.com/story/show/154329640)

## What

We want to take advantage of the billing API's price calculation for estimating costs, so we don't duplicate any work/effort on rebuilding the pricing formulas.

We decided add a new endpoint `POST /report/:org_uid?from=...&to=...` which would accept a list of fake resource events and retrieve pricing information for them as a report.

Specifically, we build a temporary table in the database with the events and use it to generate the report.

## Dependencies

 - PR in paas-cf https://github.com/alphagov/paas-cf/pull/1231

## How to review

 * Code review
 * All tests shall pass
 * You can simulate events with: 
```
# Start the server in development. Follow instructions in the README.md to update the right user.
make run-dev  DEPLOY_ENV=hector

# you must login into CF dev
cf login ... 

# Push some plans as the ones defined in prod
COLLECTOR_URL=http://localhost:8881 ./scripts/push-pricing.sh update scripts/pricing_plans.json 

# Generate a fake report
from="2018-01-01T00:00:00Z"
to="2018-01-31T00:00:00Z"
org="fake-org"

curl -X POST \
    -v \
    -qs "http://localhost:8881/report/${org}?from=${from}&to=${to}"  \
    -H "Content-Type: application/json" \
    -H "Accept: application/json" \
    -d @- <<EOF | jq .
{
  "events": [
    {
        "name": "o1s1-app1",
        "space_guid": "o1s1",
        "plan_guid": "1bc93270-3d89-4ddd-a916-53e1b1d899e9",
        "memory_in_mb": 1
    },
    {
        "name": "o1s1-db1",
        "space_guid": "o1s1",
        "plan_guid": "31bb8b32-d5e0-4310-8a8d-3e8bff4b2bbc"
    }
  ]
}
EOF
```

## Who can review

Anyone but @keymon or @dcarley 


server/server.go
```diff
@@ -27,38 +27,39 @@ func New(db db.SQLClient, authority auth.Authenticator, cf cloudfoundry.Client)
 	// Never crash on panic
 	e.Use(middleware.Recover())
 
-	// Require a token for all requests
-	e.Use(auth.UAATokenAuthentication(authority))
-
 	// Validate and parse range query param
 	e.Use(api.ValidateRangeParams)
 
+	authGroup := e.Group("", auth.UAATokenAuthentication(authority))
```
chrisfarms: These `Group` are much prettier 👍

api/auth_test.go
```diff
@@ -26,14 +26,15 @@ func (sa *SimpleAuthorizer) Admin() bool {
 type SimpleAuthenticator struct {
 	admin                bool
 	authorizedSpaceGUIDs []string
+	authorizationError   error
 }
 
 func (sa *SimpleAuthenticator) Authorize(c echo.Context) error {
-	return nil
+	return sa.authorizationError
```
chrisfarms: 👍

db/db.go
```diff
@@ -26,32 +27,86 @@ type SQLClient interface {
...
 	pc := &PostgresClient{
-		Conn: conn,
+		db:   db,
+		tx:   nil,
+		Conn: db,
 	}
 	return pc, nil
 }
 
+func (pc *PostgresClient) BeginTx() (SQLClient, error) {
```
chrisfarms: PostgresClient is becoming a God Objecty ... is it a Tx ... it's a DB

Think this is a code smell that the "methods" on this type should just be utility functions or something.

keymon:
> It makes me sad that we don't have easy access to the monetized_resources logic directly from the database any more. 

The problem is that there is not a practical way to parametrise a table in a SQL query within a function, without making a complete mess.

db/db.go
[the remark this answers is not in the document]
keymon: The temporary tables are only accessible by the same session or transaction, but our implementation of SQLClient encapsulates the `sql.DB`. 

The alternative is that SQLClient does not have the `sql.DB` but receives always a `sql.Tx` (or the SQLConn interface) for all the parameters, and the caller must pass the right `sql.DB` or create the `db.Tx`.

chrisfarms:
> The problem is that there is not a practical way to parametrise a table in a SQL query within a function, without making a complete mess.

Understood 🙌 👍 😢

api/api.go
```diff
@@ -342,21 +392,102 @@ func authorizedSpaceFilter(authorizer auth.Authorizer, rng RangeParams, sql stri
...
+				) as price
+			from
+				authorized_resources b
+			inner join
+				valid_pricing_plans vpp
+			on b.plan_guid = vpp.plan_guid
+				 and vpp.valid_for && b.duration
+				 and vpp.valid_for && tstzrange( '{{ .Range.From }}', '{{ .Range.To }}' )
+			where
+				b.duration &&  tstzrange( '{{ .Range.From }}', '{{ .Range.To }}' )
```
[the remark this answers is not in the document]
keymon: You are completely right here.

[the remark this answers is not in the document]
keymon: Added a commit to fix that

keymon:
@chrisfarms Please have a look, we removed that string interpolation.

server/server.go
keymon: I changed that to put the routes directly in the echo middleware.

chrisfarms: Looking at the code I think that `UAATokenAuthentication` middleware needs to match on `c.Request().URL.Path` not `c.Path()`

keymon: BTW, I think we do not have tests for this? shall we add tests?

chrisfarms: Would be nice to avoid these regressions, however the UI and browser oauth-flow _should_ be temporary in paas-billing ... the end game is that the UI to view usage/billing would be in paas-admin and paas-billing will just be a JSON api and won't need to care about login and we can drop this code.

server/server.go
```diff
@@ -27,38 +27,42 @@ func New(db db.SQLClient, authority auth.Authenticator, cf cloudfoundry.Client)
...
+	adminGroup := authGroup.Group("", auth.AdminOnly)
+
+	e.GET("/oauth/authorize", authority.Authorize)
+	e.GET("/oauth/callback", authority.Exchange)
+
+	e.POST("/report/:org_guid", api.NewSimulatedReportHandler(db))
```
chrisfarms: One last thing 😇 

I'm not a fan of the name clash between the `POST /report/:org_guid` route and the `GET /report/:org_guid`.

I think we should prefix our 'estimatation' endpoints with something to make it clear that it is a unique feature to do with predicting costs and we can probably safely drop the :org_guid as it doesn't add much value.

I think `e.POST("/forcast/report", NewSimulatedReportHandler)` would be good.

keymon: Yes sir!

keymon:
@chrisfarms I was wondering if you wanted us to rebase.... but I see that not :-D

PR #14, opened 2018-02-20
[#154395415] Refresh concurrently, refresh only on schedule

camelpunch:
## What

Allow the temporary view of the billing data to be refreshed concurrently with other operations. Remove code that triggered the view to be refreshed on pricing plan changes.

In order to add the concurrent behaviour, we [had to](https://www.postgresql.org/docs/9.6/static/sql-refreshmaterializedview.html) add a unique index to the view.

## How to review

Run the app, preferably with some production-like data in the local database. Ensure that the new schema is applied successfully.

### Dump production billing data

```
./scripts/cf_subshell_scoped_login.sh prod
cf target -o admin -s billing
cf conduit billing-db -- \
  docker run -ti \
    -e PGDATABASE \
    -e PGUSER \
    -e PGPASSWORD \
    -e PGPORT \
    -e PGHOST=docker.for.mac.localhost \
    --name billing-dump \
    postgres:9.6-alpine \
    pg_dump -Fc -f billing.dump
docker cp billing-dump:billing.dump .
docker rm -f billing-dump
```

### Start the database server

```
docker run --name billing -e POSTGRES_PASSWORD= -p 5432:5432 -d postgres:9.6-alpine
```

### Import the billing data dump
```
docker cp billing.dump billing:/
docker exec -ti billing pg_restore --create --clean --no-privileges --no-owner -f billing.sql billing.dump
docker cp billing:/billing.sql .
vi billing.sql # remove reassign_owned() references
docker cp billing.sql billing:/
docker exec -ti billing sh -c 'psql -U postgres < billing.sql'
```

### Run the billing app locally
```
DATABASE_URL=postgres://postgres:@localhost:5432/rdsbroker_b946a69b_76f4_4cbc_80df_5cee2d54b7ad?sslmode=disable make run-dev
```

Trigger a view refresh. Restarting the app should do this. Ensure there are no errors in the logs, and ideally ensure that the data has changed in the billable view.

## Merging

Once this PR is merged, please update the temp commit on https://github.com/alphagov/paas-cf/pull/1239 with the tag value from CI and merge to release to production.

## Who can review

Not @camelpunch or @LeePorte


db/sql/030_create_billable.sql
```diff
@@ -74,13 +74,14 @@ event_ranges as (
 -- this results in a row per-instance
 resources as (
 	select
-		t.*,
-		generate_series(1, t.inst_count)
+		t.id || '-' || t.guid || '-' || generate_series(1, t.inst_count) AS billable_id,
```
chrisfarms: Do we really need to introduce a new column here? Is it not possible to keep the unique column as `id` rather than `billable_id`

henrytk:
In the message for the commit where we stop updating the view after changing pricing plans it says:

> This change is in order to remove the deadlock bug that is currently present in the billing app.

I don't really understand how this deadlock is happening. Is this a result of multiple view updates concurrently, view updates conflicting with other operations, or something else?

henrytk:
Question for the audience: how can I tell if a database table is currently locked, or how many rows of a table are currently locked?

LeePorte:
The deadlock was occurring on multiple view updates and on other operations such as taking a dump for working locally.

LeePorte:
https://wiki.postgresql.org/wiki/Lock_Monitoring

db/bindata.go
```diff
@@ -90,7 +90,7 @@ func sql001_create_to_seconds_functionSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "sql/001_create_to_seconds_function.sql", size: 230, mode: os.FileMode(420), modTime: time.Unix(1515665547, 0)}
+	info := bindataFileInfo{name: "sql/001_create_to_seconds_function.sql", size: 230, mode: os.FileMode(436), modTime: time.Unix(1519124886, 0)}
```
henrytk: Why is the file mode changing?

chrisfarms: I assume this was change was generated by go-bindata. So this isn't even a _real_ file. Very odd that you would want the fake file permission of group execute for a fake file! 

We have a story to remove the use of go-bindata since there are some trust issues with the repository's owner.

camelpunch: Yeah it's probably part of an elaborate hack.

PR #15, opened 2018-02-23
[#154916385] Introduce pricing components

bandesz:
## What

This change removes the formula from the pricing_plans table and introduces the pricing_plan_components table where you can break down a plan into multiple formulas. This will help us to separate e.g. a memory usage cost from a disk usage cost.

None of the API endpoints have been changed in this story to return with the detailed breakdown, so everything is still summed up at least to the plan level.

Additional improvements:
 * Return 404 and a JSON error message if a single record can not be found. (Originally this was 200)
 * Use the same Plan fixture structs in the API tests
 * Get rid of the /usage endpoint as it is deprecated
 * Remove the pricing_plans UI
 * Update plan updater script to handle components

## How to review

Mostly code review.

After deploying the paas-billing app to your dev env you can test the plan updater script.

```
BRANCH=pricing_components_154916385 ENABLE_BILLING_APP=true SELF_UPDATE_PIPELINE=false make dev pipelines
```

If you already inserted some rows into the pricing_plans or pricing_plan_components table then clean them up with:

```
truncate pricing_plans cascade;
alter sequence pricing_plans_id_seq restart;
alter sequence pricing_plan_components_id_seq restart;
```

To test create and update:
```
COLLECTOR_URL=https://paas-billing.${DEPLOY_ENV}.dev.cloudpipelineapps.digital ./push-pricing.sh create pricing_plans.json pricing_plan_components.json
COLLECTOR_URL=https://paas-billing.${DEPLOY_ENV}.dev.cloudpipelineapps.digital ./push-pricing.sh update pricing_plans.json pricing_plan_components.json
```

To test download:
```
COLLECTOR_URL=https://paas-billing.${DEPLOY_ENV}.dev.cloudpipelineapps.digital ./push-pricing.sh get pricing_plans.json pricing_plan_components.json
```

## Who can review

Not me.


api/api.go
```diff
@@ -325,41 +303,78 @@ func ListEventUsageForResource(db db.SQLClient) echo.HandlerFunc {
...
+				iso8601(lower(mr.duration)) as from,
+				iso8601(upper(mr.duration)) as to,
+				sr.price_in_pence::bigint
 			from
-				monetized_resources
-			where
-				guid = $1
+				summed_resources sr
+			inner join
+				monetized_resources mr
```
chrisfarms: These queries basically joining on to another version of themselves looks a bit odd to me ... it took me quite while to grok it.

could we not simplify them to remove the join something like this:

```
...
                return withAuthorizedResources(Many, billableViewName, c, db, `
                        select
                                guid,
                                org_guid,
                                space_guid,
                                pricing_plan_id,
                                pricing_plan_name,
                                iso8601(lower(duration)) as from,
                                iso8601(upper(duration)) as to,
                                sum((price * 100)::bigint) as price_in_pence
                        from
                                monetized_resources
                        where
                                guid = $1
                        group by
                                guid, org_guid, space_guid, pricing_plan_id, pricing_plan_name, duration
                        order by
                                pricing_plan_id
                `, resourceGUID)
...
```

bandesz: We could but I wanted to avoid it as it looks hacky and it's not clear anymore what are the real group by fields. If you think that's still better/clearer in the end I'm happy to change them.

bandesz: I added an extra commit to change these two queries.

chrisfarms: nice ... yeah I think that's much clearer.

PR #17, opened 2018-03-02
[#155130407] Replace bindata

chrisfarms:
## What

We are removing our dependency on go-bindata that was used to bundle SQL files with our binary, and replacing it with reading the source files from the disk (since they are copied along with our binary anyway).

By default it will look for the schema files in`$PWD/db/sql/*.sql` which should work both locally on any shell that provides the `$PWD` environment variable (most) and in cloudfoundry which sets the `$PWD` to the correct path where the buildpack ran.

The default behaviour can also be overridden using a new `DATABASE_MIGRATION_DIR` environment variable in case these do not work in the environment you are using.

## How to review

* can you still run the tests?
* can does it still work in [cf](https://github.com/alphagov/paas-cf/pull/1260)?

## Who can review

Not @chrisfarms


db/migrate.go
```diff
@@ -2,17 +2,26 @@ package db
...
 	"github.com/lib/pq"
 )
 
 // InitSchema initialises the database tables and functions
-//go:generate go-bindata -pkg db -o bindata.go sql/...
 func (pc *PostgresClient) InitSchema() (err error) {
-	names := AssetNames()
-	sort.Strings(names)
-
+	// WARNING: this won't work if the binary is shipped without the code.
```
camelpunch: I don't think this is true. If someone wanted to, they could deploy this software with their own SQL dir.

Not suggesting we should do it, but hypothetically wouldn't it be possible to separate the SQL dir and ship with just binary + SQL?

Can we delete this line?

dcarley: 👍 

The reasons were stronger when it was using `runtime.Caller` because it referred to absolute paths on the machine that built the binary, which could have been very confusing. I don't think it's worth keeping now.

chrisfarms: removed

PR #18, opened 2018-03-02
[#154835556] Compose billing events

bandesz:
## What

We want to collect the Compose scaling events for billing purposes.

The Compose API provides the audit events through the /audit_events endpoint [1]. The API lists the events in chronological descending order and we can use the cursor parameter for proper pagination. This behaviour is different from the Cloud Foundry usage events APIs which lists the events in ascending order and makes the event fetching trivial.

In this case the fetching algorithm is the following:
 - we store the latest event id in the compose_audit_events_cursor table with
   the "latest_event_id" key
 - we store the cursor (last event's id on the current page) if we got a full
   page back from the API
 - if the stored cursor is nil we fetch the latest events and we update the
   latest_event_id in the database
 - if the stored cursor is not nil we fetch the events after that cursor (next
   page) and we don't touch the latest_event_id
 - when we get the next batch of data we check if we reached the
   latest_event_id (or there is no more events) and we stop there
 - we save only specific events, where event.Event is "deployment.scale.members"

The original collector was refactored as originally it was running to collection jobs which made extending it and testing it hard. Now the collector only runs one task.

The default schedule was changed to 15 minutes as there is no need to hit the APIs every minute.

[1] https://apidocs.compose.com/reference#2016-07-get-audit_events

## How to review

Code review.

Deploy and test it using the instructions from https://github.com/alphagov/paas-cf/pull/1265

## Who can review

Not @LeePorte or me.


db/sql/040_create_componse_events.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE IF NOT EXISTS compose_audit_events (
+	id SERIAL,
+	event_id CHAR(24) UNIQUE,
+	created_at TIMESTAMP,
```
henrytk: Should this have a timezone?

bandesz: Although it seems the API returns the event in GMT it won't hurt, so I changed it to "TIMESTAMPTZ"

henrytk:
This is awesome. Merging.

PR #20, opened 2018-03-07
[155326674] Add VAT rates

bandesz:
## What

Add VAT calculations to billing reports.

* create the vat_rates table
* add a vat_rate_id mandatory field to the pricing_plan_components
* populate the vat_rates table with two default VAT rates: standard and zero rate
* modify the queries to return the price including and excluding VAT
    * note: the vat_rate_name won't be returned in any of the top level queries as we group by the plan.
* add API endpoints to handle VAT rate

## How to review

Code review.

Deploy it following the instructions from https://github.com/alphagov/paas-cf/pull/1269. 

## Who can review

Not me.


api/api.go
```diff
@@ -337,7 +347,8 @@ func ListEventUsage(db db.SQLClient) echo.HandlerFunc {
 				memory_in_mb,
 				iso8601(lower(duration)) as from,
 				iso8601(upper(duration)) as to,
-				sum(price::bigint) as price
+				sum(price_inc_vat::bigint) as price_inc_vat,
```
chrisfarms: Clearly you didn't introduce this...but..

[One of these things is not like the other](https://www.youtube.com/watch?v=ueZ6tvqhk8U) ... pretty sure it's a bug/omission that `ListEventUsage` is the only endpoint to return price not "in pence" and just truncating the decimal!

However I think the x100 should generally live in the monitize_resources query (if anywhere) so I'm making a note to clean this up as part of the story that decides what endpoints we really need.... so don't expect this fixed now.

db/sql/045_create_vat_rates.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS vat_rates (
+	id SERIAL PRIMARY KEY,
+	name VARCHAR(32),
+	rate NUMERIC,
+
+	CHECK (length(trim(name)) > 0),
+	CHECK (rate >= 0)
+);
+CREATE INDEX IF NOT EXISTS idx_vat_rates_id ON vat_rates (id);
```
chrisfarms: `PRIMARY KEY` will create an index so I think this extra index is superfluous

bandesz: fixed

api/api.go
bandesz: Yeah, was wondering about this myself. I'm fine with changing these in the follow-up story.

db/sql/045_create_vat_rates.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS vat_rates (
+	id SERIAL PRIMARY KEY,
+	name VARCHAR(32),
+	rate NUMERIC,
+
+	CHECK (length(trim(name)) > 0),
+	CHECK (rate >= 0)
+);
+
+INSERT INTO
```
keymon: This adds rows every time the app is restarted. See https://github.com/alphagov/paas-billing/pull/21#issuecomment-372408162

PR #22, opened 2018-03-16
[#154835806] Expose memory and storage in MB for services

henrytk:
## What

This enables memory and storage in MB data to be stored with the pricing plans and passed into pricing plan components for calculating bills.

## How to review

* Code review.
* Make sure tests pass

It is not clear what our process is for adding and testing new SQL files, so we've added tests for applying new 'migrations' files. Personally, I think we should always add new SQL files (only editing ephemeral structures, such as views; or non-data objects, such as functions) by giving them a file name which is higher alphanumerically.

## Who can review

Anyone but me, @46bit or @LeePorte


henrytk:
@samcrang it looks like in https://github.com/alphagov/paas-billing/pull/21 a [currency column was added](https://github.com/alphagov/paas-billing/blob/3d00d4f6f095fecf0056e35544d7a50c042339ec/db/sql/049_modify_pricing_plan_components_currency.sql#L5), but the JSON scripts weren't updated.

henrytk:
I've pushed a commit which fixes the issue with the JSON files. See commit message for details.

keymon:
I love the idea of testing the migrations. 

Additionally, I think we should soon implement a better migration mechanism, that can keep track of the migrations and ensure that each migration is runs once. We are adding too much complexity to ensure the idem-potentially of each .sql script, and that is difficult to test and maintain.

fixtures/plans.go
```diff
@@ -17,23 +17,27 @@ type PricingPlanComponent struct {
...
-	PlanGuid   string
-	ValidFrom  time.Time
-	Components []PricingPlanComponent
+	ID          int
+	Name        string
+	PlanGuid    string
+	ValidFrom   time.Time
+	Components  []PricingPlanComponent
+	MemoryInMb  uint
+	StorageInMb uint
```
keymon: I wonder if one option could be add the `MemoryInMb` and `StorageInMb` to each `PricingPlanComponent` instead to the plan. That would make it easier to later split the usage of a service, for instance: S3 storage usage vs. number of API calls or Bandwidth consumed.

But we do not have this use case right now.

PR #23, opened 2018-03-29
[#155326294] Declarative configuration refactoring

chrisfarms:
## What

We want:

* to pull together all the recent work to allow us to generate the BillableEvents with pricing data that we require 
* to remove any endpoints / code we do not need to make maintence easier.
* to change the architecture for this application in a way that allows us to declaratively configure our pricing data (ie from our deployment manifest) by specifying a config.json file
* to make the tests comprehensible so changes no longer cause huge cascades of failures
* to have a valid configuration for our production pricing plans (which should be moved to our manifest in the future) 

To address these:

* We have dropped all the existing UI and API code and replaced it with a single path `GetBillableEvents` which returns processed events with pricing data.
* We have dropped the "fixtures" and heavy use of Before/After filters style of tests that was getting too difficult to reason about and replaced it with a series of flatter more repeated tests with diagrams to help people understand the intention.
* 

Along the way:

* A bug was found that caused incorrect (null) outputs due to events of zero duration ... this was resolved by excluding events of zero duration and adding constraints so that it cannot occur
* An issue was found where many short lived events would round to zero. For now we have agreed to hard code a minium event cost of 1p ... however we may need to look at this in the future or at least allow configuration via the config file
* We were previously outputting prices in a mixture of float or in integer-pence ... neither of these were ideal so it has been switched to string representations.
* Performance of generating events was poor (50+ seconds before for large orgs for a month) ... by moving more of the processing into the periodic reporting tasks and excluding some of the junk data (CATS/PERF/ACC etc) this came down to 1s

## How to review

## Who can review


collector/cloudfoundry/event_fetcher.go
```diff
@@ -40,14 +47,65 @@ func (e *EventFetcher) FetchEvents(logger lager.Logger, fetchLimit int, recordMi
...
 func (e *EventFetcher) Name() string {
 	return fmt.Sprintf("cf-%s-usage-event-collector", e.client.Type())
 }
+
+func (e *EventFetcher) InsertEvents(tx *sql.Tx, batch *cloudfoundry.UsageEventList) error {
+	valueStrings := make([]string, 0, len(batch.Resources))
+	valueArgs := make([]interface{}, 0, len(batch.Resources)*3)
+	i := 1
+	for _, event := range batch.Resources {
+		// if event.MetaData.GUID == "" {
```
LeePorte: This block shouldn't be commented out

reporter/currency_conversion_test.go
```diff
@@ -0,0 +1,386 @@
+package reporter_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	// . "github.com/onsi/gomega"
+)
+
+var _ = Describe("Currency Conversion", func() {
+	// var (
```
LeePorte: This test is commented out?

PR #36, opened 2018-06-15
[#157438891] Historical service data

LeePorte:
What
----

This commit adds the new `cfstore` package to the billing app. The
purpose of this package is to enable the collection of service and
service plan data from Cloud Foundry and storing this data within the
billing database.

The consumers of this are the pricing calculator and the billing app
itself. Due to the names of the services and plans being used
as a method of identifying which plan to use, this was clearly a fragile
approach and when a plan was renamed the result would be the GUID being
displayed which has very little meaning to our users.

In order to aid local testing two make targets have been added, these
start the docker postgres container and clean up the resulting
container.

How to review
-----

- Code review
- See in action

Who can review
-----

Neither @chrisfarms nor @LeePorte


cfstore/cfstore.go
```diff
@@ -0,0 +1,204 @@
...
+		var serviceValidFrom *time.Time
+		err = tx.QueryRow(`
+			select valid_from
+			from services
+			where guid = $1
+			order by valid_from desc
+			limit 1
+		`, plan.ServiceGuid).Scan(&serviceValidFrom)
+		if err == sql.ErrNoRows {
+			panic("meh dunno what to do")
```
paroxp: Do we want to panic here? Can we not:

```
if err != nil {
  return err
}
```

for the error to be handled properly later on and not kill the application?

cfstore/cfstore_client.go
```diff
@@ -0,0 +1,132 @@
...
+			servicePlan.Free = servicePlanResource.Entity.Free
+			servicePlan.Name = servicePlanResource.Entity.Name
+			servicePlan.Description = servicePlanResource.Entity.Description
+			servicePlan.ServiceGuid = servicePlanResource.Entity.ServiceGuid
+			servicePlan.Extra = servicePlanResource.Entity.Extra
+			servicePlan.UniqueId = servicePlanResource.Entity.UniqueId
+			servicePlan.Public = servicePlanResource.Entity.Public
+			servicePlan.Active = servicePlanResource.Entity.Active
+			servicePlan.Bindable = servicePlanResource.Entity.Bindable
+			servicePlans = append(servicePlans, servicePlan)
```
paroxp: Not a blocker, but you can also do:

```go
servicePlans = append(servicePlans, ServicePlan{
  Guid: servicePlanResource.Meta.Guid,
  CreatedAt: servicePlanResource.Meta.CreatedAt,
  UpdatedAt: servicePlanResource.Meta.UpdatedAt,
  ...
})
```

personally find it cleaner.

cfstore/cfstore_client.go
```diff
@@ -0,0 +1,132 @@
...
+			service.CreatedAt = serviceResource.Meta.CreatedAt
+			service.UpdatedAt = serviceResource.Meta.UpdatedAt
+			service.Label = serviceResource.Entity.Label
+			service.Description = serviceResource.Entity.Description
+			service.Active = serviceResource.Entity.Active
+			service.Bindable = serviceResource.Entity.Bindable
+			service.PlanUpdateable = serviceResource.Entity.PlanUpdateable
+			service.ServiceBrokerGuid = serviceResource.Entity.ServiceBrokerGuid
+			service.Tags = serviceResource.Entity.Tags
+			services = append(services, service)
```
paroxp: Same as line 84

cfstore/cfstore.go
chrisfarms: yep def shouldn't panic that shouldn't have been left in

cfstore/cfstore_client.go
paroxp: Instead of

```go
servicePlan := ServicePlan{}
servicePlan.Guid = servicePlanResource.Meta.Guid
servicePlan.CreatedAt = servicePlanResource.Meta.CreatedAt
servicePlan.UpdatedAt = servicePlanResource.Meta.UpdatedAt
servicePlans = append(servicePlans, servicePlan)
```

Wasn't clear enough. Sorry!

PR #37, opened 2018-06-15
[#157307645] Authorize access to billing data based on Managers role

paroxp:
What
----

We'd like to make sure the user has access to the organisations they'd
like to see billing for. We're refactoring the interface and applying
some small changes through out the module to satisfy that need.

We've got a need to extract the token in more places now. It has been
previously done as part of the getVerifiedScopes() method, where it
didn't actually save the outcome for the future to come.

We've moved it outside of that method and saving the state in the
authorizer for other uses. If the method is called twice in the raw,
it'd reuse already prepared claims, in order to save some time and
resources.

We'd like to check if the user has either billing or org manager role in
requested organisation before they receive the information. In order to
achieve that, we're asking the CF API to provide us with the
organistations for that user where they have the desired role, and then
iterating over them to see if the requested guids are in fact listed.

We're removing the Spaces functionality, as in our belief, it is no
longer required nor used.

How to review
-----

- Code review + Sanity check
- Deploy to your environment - use the X
- Merge before X


eventserver/auth/authorizer.go
```diff
@@ -1,6 +1,6 @@
 package auth
 
 type Authorizer interface {
-	Spaces() ([]string, error)
 	Admin() (bool, error)
+	Organisations([]string) (bool, error)
```
keymon: I would rename this interface method because the actual meaning that we are giving to it is: "Has this user permissions to get billing information for this set of orgs?"

For instance you could call it `BillingAllowedForOrgs(orgs)`

eventserver/server_auth.go
```diff
@@ -9,12 +9,15 @@ import (
 
 var (
 	ErrInvalidCredentials = errors.New("invalid credentials")
-	ErrAdminOnly          = errors.New("billing data currently requires cloud_controller.admin or cloud_controller.global_auditor scope")
+	ErrAdminOnly          = errors.New("you need to be billing_manager or an administrator to retreive the billing data")
```
keymon: I would rename the error variable to `ErrBillingPermissions` or something else. It is not for "Admins only" anymore

eventserver/server_billable_handler_test.go
```diff
@@ -96,9 +96,10 @@ var _ = Describe("BillableEventsHandler", func() {
 		Expect(res.Header().Get("Content-Type")).To(Equal("application/json; charset=UTF-8"))
 	})
 
-	It("should require admin scope (for now)", func() {
+	It("should require admin scope or manager role", func() {
```
keymon: Actually the test check "admin scope or billing access permissions" or  "admin scope or billing access permissions or org manager"

eventserver/auth/uaa_authorizer.go
```diff
@@ -69,20 +70,36 @@ func (a ClientAuthorizer) client() (*cfclient.Client, error) {
 	})
 }
 
-func (a *ClientAuthorizer) Spaces() ([]string, error) {
+func (a *ClientAuthorizer) Organisations(requestedOrgs []string) (bool, error) {
```
keymon: I am not really convinced that we are adding here some functionality but not really testing it, because in the tests we mock these methods. 

Maybe we should refactor the tests to actually mock the cf client itself or use a fake api server?  

With this change we add more logic than before, because we have multiple branches depending the scope (admin), if the user is a OrgManager or if the user is BillingManager. Before it was only returning the list of space IDs one to one.

Anyway, if you are OK with it, won't block this PR because this because there was not a test before.

eventserver/server_auth.go
paroxp: Nice touch.

eventserver/server_billable_handler_test.go
paroxp: Maybe. I though of it like `The user has either (billing or organisation) manager role`.

keymon: sure, I meant make it more meaningful

keymon:
@paroxp how can I test a request ? do you have a curl ?

paroxp:
I've used the curl mentioned in the readme: https://github.com/alphagov/paas-billing#get-usage_events

eventserver/auth/authorizer.go
keymon: I would rather change the name, because for me it lacks of meaning as it is. As a developer if I have to look at whatever function calls this, I would have to look at the implementation to know what is doing.

It should say something like:

 - `HasBillingAccess(orgs)`
 - `IsBillingManager(orgs)`
 - `BillingAllowed(orgs)`
 - `BillingManager(orgs)`

eventserver/auth/uaa_authorizer.go
keymon: OK, but as said: The previous implementation was simply forwarding values, in the new  one we add a more complex behaviour. Proof of it was that the original implementation had a bug.

PR #39, opened 2018-07-16
WIP: [#SUPPORT] Log error if authentication failed

keymon:
What
----

If there is any error during credential verification, we log it
so we can troubleshoot it as needed.

How to review
-----

Code review

Who can review
-----

not me


keymon:
Actually there is a more deep bug here... auth fails if there are multiple keys in UAA. I will do a new PR.

PR #42, opened 2018-07-20
[#158943954] Add Aiven Elasticsearch pricing

LeePorte:
What
----

We would like to start charging for Aiven Elasticsearch v5/v6 clusters
when the service becomes available in the PaaS marketplace. This PR
informs the billing app about the 2 different plans we will initially
export.

How to review
-----

Code review

Who can review
-----

Not @LeePorte or @jpluscplusm

### DIFF
--- a/config.json
+++ b/config.json
@@ -903,6 +903,32 @@
 					"vat_code": "Standard"
 				}
 			]
+		},
+		{
+			"name": " small-ha-es5",
+			"valid_from": "2017-01-01",
+			"plan_guid": "6c35a479-9622-4852-b80e-b21b40e0378f",
+			"components": [
+				{
+					"name": "instance",
+					"formula": "ceil($time_in_seconds/3600) * 0.548",
+					"currency_code": "USD",
+					"vat_code": "Standard"
+				}
+			]
+		},
+		{
+			"name": " small-ha-es6",
+			"valid_from": "2017-01-01",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -903,6 +903,32 @@
 					"vat_code": "Standard"
 				}
 			]
+		},
+		{
+			"name": "elasticsearch v5 small ha (aiven)",
```
chrisfarms: these names should probably follow the pattern from [ADR025](https://team-manual.cloud.service.gov.uk/architecture_decision_records/ADR025-service-plan-naming-conventions/) prefixed with the service name, and match the marketplace names.

The `(aiven)` is not required, that was added to the compose was as part of the decomissioning of those services.

I think they should probably be:

`elasticsearch small-ha-es5` and `elasticsearch small-ha-es6`

I'm also a bit concerned that we are using `es6` rather than semver-style `6.3` as per the ADR? ... as this might pose problems if we need to offer both versions (not something for this PR, since that's in paas-cf land, but we might want to addresses before we enable the service)

LeePorte: I'll update the names.

There is only the option for the latest in either the ES5 or ES6 series from Aiven at point of instance creation.

chrisfarms: Okay ... I still think the "es" is redundent tho ... it would be like "postrges small-pg9.5" ... I'll create a story to bring up at planning

PR #45, opened 2018-08-01
[#157205103] Store org and space name from historic data

henrytk:
What
----

We are going to start using historical org and space name data from the CF API to fill in names where a billed tenant would otherwise see just a GUID. This pull request just makes the data available, but doesn't yet affect visibility of the org/space names.

How to review
-----

The usual: code review and make sure all tests pass.

Because we are not doing the tenant-visible side yet it will be hard to test whether this really provides what is required, but I suggest we defer such judgement until we actually start work on the front end side.

Who can review
-----

Anyone but me or @jpluscplusm


Makefile
```diff
@@ -22,7 +22,7 @@ run-dev: bin/paas-billing
 test: fakes/fake_usage_api_client.go fakes/fake_cf_client.go fakes/fake_event_fetcher.go fakes/fake_compose_client.go fakes/fake_event_store.go fakes/fake_authorizer.go fakes/fake_authenticator.go fakes/fake_billable_event_rows.go fakes/fake_usage_event_rows.go fakes/fake_cf_data_client.go
 	$(eval export TEST_DATABASE_URL=${TEST_DATABASE_URL})
 	$(eval export APP_ROOT=${APP_ROOT})
-	ginkgo -nodes=8 -r
+	ginkgo $(ACTION) -nodes=8 -r $(PACKAGE)
```
chrisfarms: 👍

cfstore/cfstore_client.go
```diff
@@ -1,46 +1,14 @@
 package cfstore
```
richardTowers: Does this file provide any value over the code just using go-cfclient directly now? Could we just remove it?

richardTowers:
Tested this by deploying billing to my local dev env (with `cf push`) and curling the events. I can also see the nice org_name and space_name fields.

Merging, then I'll bump it in to paas-cf.

cfstore/cfstore_client.go
henrytk: Unfortunately I don't think the go-cfclient library offers an interface for the client that we can create a fake version of. Being able to fake this client provides a convenient and idiomatic way of testing when certain responses come back from the CF API.

alext: It's become an emerging idiom in this sort of context to declare the interface at the point that it's consumed. As things stand, there shouldn't be a need for the `Client` type because `*cfclient.Client` itself already implements the defined interface.

This means that the consuming code can define an interface that it accepts, and we can then pass in either an instance of `*cfclient.Client`, or a fake that also implements the interface.

PR #49, opened 2018-08-30
[#158665945] Use plans `unique_id` to identify the pricing_plans 

LeePorte:
What
----

This PR changes how we map the service plans for billing purposes with the events. 

Currently we were using the Cloud Foundry generated `plan_guid` for this purpose. This had the limitation that different configurations of the service plans would be needed on a per environment basis.

Since we did the work to collect the historic data in https://github.com/alphagov/paas-billing/pull/36 we have the broker reported plan unique IDs available for this purpose.

This PR would leverage the table `service_plans` to map the events to the corresponding plans using the plan unique ID reported by the broker. It also updates the embedded `config.json` to use the `unique_id` instead of the specific CF plan id.

Notes about the implementation: 

 - Old Pricing Calculator URLs would break:
    - We change the `config.json` IDs, which means that the `/pricing_plans` would report a different GUID for the plans.
   - The forecast events would use the `unique_id`. This makes sense as the `/pricing_plans` where `paas-admin` calculator gets the values would be using `uinque_ids`.
 
 - We add some additional tests to cover this case
 - We add logic to map any service not stored in the historic tables `services` and `service_plans` to a "Unknown Plan" with cost 0.
 - We added a SQL script to add a missing service plan for `redis tiny (compose)`.

How to review
-----

- Code review
- Deploy https://github.com/alphagov/paas-cf/pull/1502
- Check you have bills for services in your dev env, admin is a good org to test with.

Optionally testing in prod data (requires access to prod): 
 - We did compare all the costs calculated with this version for all the orgs in prod.
 - All the notes are in the private HackMD of the story. The scripts are documented in a comment below.: 

Before merge
----------------

It is important to run the script: `scripts/create_adhoc_redis_service.sql` in prod before deploying in prod:

```
cf target -o admin -s billing
cf conduit billing-db -- psql 
```
and paste the script.

Who can review
-----

Not @LeePorte or @keymon


keymon:
# Verify with prod data 

Import the DB following the instructions in https://github.com/alphagov/paas-billing/pull/14


Start the service locally 

```bash

# Load AWS creds
awssts sts:hectorrivasgandara@gov-paas-dev

# login in CF

PROD_DB_NAME=....

DATABASE_URL=postgres://postgres:@localhost:5432/${PROD_DB_NAME}?sslmode=disable make run-dev DEPLOY_ENV=rafalp

```

Run the comparision

```bash

cf oauth-token > /tmp/prod-token
cf curl  '/v2/organizations?results-per-page=100&q=status:active'| jq -r '.resources[].metadata.guid' > /tmp/prod-orgs.txt

cf login -a https://api.leeporte.dev.cloudpipeline.digital -u admin
cf oauth-token > /tmp/dev-token



TMP_DIR=/tmp/billing-check
mkdir -p $TMP_DIR

get_report() {
    token=$1
    env=$2
    url=$3
    org=$4
    curl -qs \
        -H "Authorization: ${token}" \
        "${url}/billable_events?org_guid=$4&range_start=2017-01-01&range_stop=2018-08-30" \
        > "${TMP_DIR}/${org}-${env}.json"
}

sum_report() {
    env=$1
    org=$2
    # The sub is to reduce the precision
    jq  -r 'map(.price.inc_vat | tonumber) | add | tostring | sub("(?<num>[0-9]+\\.[0-9]{4}).*"; .num)' \
        < "${TMP_DIR}/${org}-${env}.json"
}

dev_url="http://localhost:8881"
prod_url="https://billing.cloud.service.gov.uk"


orgs=""

for o in ${orgs}; do 
    echo -n "$o: "
    get_report "$(</tmp/dev-token)" dev "${dev_url}" "$o"
    get_report "$(</tmp/prod-token)" prod "${prod_url}" "$o"
    if diff <(sum_report prod $o) <(sum_report dev $o) > /dev/null;then
        echo "ok"
    else 
        echo "differs"
        diff <(sum_report prod $o) <(sum_report dev $o)
    fi
done
for o in ${orgs}; do
    if ! diff <(sum_report prod $o) <(sum_report dev $o);then
        echo "$o differs"
    fi
done
```

eventio/event_usage.go
```diff
@@ -17,6 +17,7 @@ type UsageEvent struct {
 	SpaceGUID     string `json:"space_guid"`
 	SpaceName     string `json:"space_name"`
 	PlanGUID      string `json:"plan_guid"`
+	PlanUniqueID  string `json:"plan_unique_id"`
```
henrytk: These names (`planGUID` and `PlanUniqueID`) are really confusing because they are both unique and they are both GUIDs.

eventstore/sql/create_billable_event_components.sql
```diff
@@ -80,7 +80,7 @@ CREATE OR REPLACE FUNCTION generate_billable_event_components() RETURNS SETOF bi
 	from
 		events ev
 	left join
-		valid_pricing_plans vpp on ev.plan_guid = vpp.plan_guid
+		valid_pricing_plans vpp on ev.plan_unique_id = vpp.plan_guid
```
henrytk: `ev.plan_unique_id` does not exist as a column on the events table. This is confusing me.

eventstore/store_billable_events_test.go
```diff
@@ -1137,4 +1167,502 @@ var _ = Describe("GetBillableEvents", func() {
...
+	         |           |                                                               .
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	 .   .   [====db1====]   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	       start       stop                                                  .
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	<=======================================PLAN1 (cf_id)===============================>.
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	-------------------------------------------------------------------------------------*/
+	It("Should include BillableEvent with event.plan_guid == pricing_plan.plan_guid", func() {
```
henrytk: This isn't helpful text. It should describe behaviour, not implementation.

eventstore/store_billable_events_test.go
```diff
@@ -1137,4 +1167,502 @@ var _ = Describe("GetBillableEvents", func() {
...
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	 .   .   [====db1====]   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	       start       stop                                                  .
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	<====================================SERVICE_PLAN (cf_id)===========================>.
+	<=======================================PLAN1 (unique_id)===========================>.
+	 .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .   .
+	-------------------------------------------------------------------------------------*/
+	It("Should include BillableEvent if event.plan_guid = service_plan.guid and service_plan.unique_id == pricing_plan.plan_guid", func() {
```
henrytk: This isn't helpful text. It should describe behaviour, not implementation.

eventio/event_usage.go
keymon: It is also confusing for me, but naming is hard. `guid` here refers to the fact that is a GUID https://betterexplained.com/articles/the-quick-guide-to-guids/. All the columns all around have that `_guid` suffix. 

But then we have two kind of `plan_guid`:
 - `plan_guid` as the GUID of the metadata reported in the events. This GUID is different across deployments, and changes if the broker is deregistered and registered again.
 -  `unique_id` reported by the CF api: https://apidocs.cloudfoundry.org/4.0.0/service_plans/list_all_service_plans.html and which is feed into by the broker. This is constant across installations.

eventstore/store_billable_events_test.go
keymon: I will think something better.  Anyway, the test is quite low level

eventio/event_usage.go
alext: How about we call it something like `plan_catalog_guid` to indicate that it comes from the service catalog returned by the broker. Or is that more confusing given the various layers of indirection here?

config.json
```diff
@@ -924,7 +924,7 @@
 		{
 			"name": "redis tiny (compose)",
 			"valid_from": "2017-01-01",
-			"plan_guid": "53ca5c56-5474-4d64-9211-fe9aee86d502",
+			"plan_guid": "1356eeeb-7c5d-4d9d-9a04-c035a2c709b3",
```
henrytk: The actual GUID is in the Git history of paas-cf:

```bash
git show b3f8018795c55f0f0bd36ed9e7b48d20bc76d87b \
  config/service-brokers/compose/catalog.json
```

It was `a8574a4b-9c6c-40ea-a0df-e9b7507948c8`, but I don't think the actual value matters anymore.

scripts/create_adhoc_redis_service.sql
```diff
@@ -0,0 +1,73 @@
+-- Add the missing historic recorded service in prod
```
henrytk: Why only in prod? I don't like adding state manually, and I don't like that it won't be in dev or staging. I think this should be codified.

keymon: If I add it to the initialisation of the app, we need to change all the tests to reflect that new thing.

That is why I did not codified automatically. 

This is to add the data once on production. The `service_plans` data is populated every now and then with data from CF and it is not added programaticaly.

henrytk: > If I add it to the initialisation of the app, we need to change all the tests to reflect that new thing.

I don't think that justifies having to manually run a script to make prod work, especially with the new added complexity of multiple prod environments. This app is complicated enough to understand without the added complexity of remembering ad hoc things we layered on top of it.

I think we should codify this.

keymon: But I insist, this is one off ever

henrytk: @keymon discussed this offline and agreed scripting against prod is okay because:

- the data is only needed in prod (specifically the existing Ireland environment). This is because of event data for a service plan exists when the service plan has been removed _before_ we started collecting historical service plan data.
- there is extra work in correcting existing tests if it is added everywhere
- this only needs to be run once

eventstore/sql/create_events.sql
```diff
@@ -274,6 +275,11 @@ INSERT INTO events with
 		coalesce(vspace.name, space_guid::text) as space_name,
 		duration,
 		plan_guid,
+		(case
+			when resource_type = 'service'
+			then coalesce(vsp.unique_id, 'd5091c33-2f9d-4b15-82dc-4ad69717fc03')::uuid
+			else plan_guid
```
henrytk: This is the last place `plan_guid` is used, so we don't need to store it in this table (line 276/7) and can remove it from all queries of event data.

keymon: I have pushed a pair of commits. This required more work than what you think, but actually allowed me to notice a existing bug.

The format of events is exposed into the forecast events endpoint, so paas-admin needs to pass the `unique_id`. I added two commits for this, one to get rid of the old `plan_guid` and other to make `forecast_events` support both

keymon: > but actually allowed me to notice a existing bug.

I was wrong there was not a bug, the feature was working fine because in the https://github.com/alphagov/paas-billing/blob/9a1c2da2de72578ee1b88a266f5ba9377e2d28bc/eventstore/store_forecast_events.go#L49-L50 we were populating the column `plan_unique_guid` with the value of `ev.PlanGUID`

But that is now the change is more explicit.

Anyway, there is still some mixed dichotomy  all around.  
 - The input config.json call this `plan_guid` when it is actually the `plan_unique_id`
 - The resulting billable_events the call the property `PlanGUID, but it is actually the value from unique.  https://github.com/alphagov/paas-billing/blob/9fcbab753142c9c373babadc24e4b8a0bc264468/eventio/event_billable.go#L56-L57

keymon:
We decided to not introduce the `plan_unique_id` but instead use  `plan_guid` all the time.

The `plan_guid` is introduced in the query that creates the events from: 
 - Hardcoded values for the case of `app`, `staging` and `task` events
 - From `service_plan_guid` before.

So **we are not changing the meaning of `plan_guid`** because previously was called `service_plang_guid` ¯\_(ツ)_/¯

I close this PR because the new implementation changes all the history.

PR #52, opened 2018-09-05
[#158665945] Use plans `unique_id` to identify the pricing_plans (attempt 2)

keymon:
What
----

This PR is the second one after closing #49

This PR changes how we map the service plans for billing purposes with the events.

Currently we were using the Cloud Foundry generated `service_plan_guid` for this purpose. This had the limitation that different CF deployments would require different pricing_plans with a different plan_guid.

Since we did the work to collect the historic data in https://github.com/alphagov/paas-billing/pull/36 we have the broker reported plan unique IDs available for this purpose.

This PR would leverage the table `service_plans` to map the events to the corresponding plans using the plan unique ID reported by the broker. It also updates the embedded `config.json` to use the `unique_id` stored in the `service_plans` table instead of the specific CF plan id.

Notes about the implementation:

 - Old Pricing Calculator URLs would break:
    - We change the `config.json` IDs, which means that the `/pricing_plans` would report a different GUID for the plans.
   - The forecast events would use this new `unique_id` when passing the value of `plan_guid`. This makes sense as the `/pricing_plans` where `paas-admin` calculator gets the values would be using `uinque_ids`.

 - We add some additional tests to cover this case
 - We add logic to map any service not stored in the historic tables `services` and `service_plans` to a "Unknown Plan" with cost 0.
 - **IMPORTANT** We added a SQL script to add a missing service plan for `redis tiny (compose)`.

How to review
-----

- Code review
- Deploy https://github.com/alphagov/paas-cf/pull/1502
- Check you have bills for services in your dev env, admin is a good org to test with.

Optionally testing in prod data (requires access to prod):
 - We did compare all the costs calculated with this version for all the orgs in prod.
 - All the notes are in the private HackMD of the story. The scripts are documented in a comment in https://github.com/alphagov/paas-billing/pull/49#issuecomment-417677182

Before merge
----------------

It is important to run the script: `scripts/create_adhoc_redis_service.sql` in prod before deploying in prod:

```
cf target -o admin -s billing
cf conduit billing-db -- psql
```
and paste the script.

Who can review
-----

Not @LeePorte or @keymon

### DIFF
--- a/config.json
+++ b/config.json
@@ -33,7 +33,7 @@
 		{
 			"name": "prometheus",
 			"valid_from": "2017-01-01",
-			"plan_guid": "09f2a841-1997-468b-be99-15020ff6c9c4",
+			"plan_guid": "b5998c91-d379-4df7-b329-11450f8459f1",
 			"components": [
 				{
 					"name": "instance",
@@ -129,7 +129,7 @@
 		{
 			"name": "postgres tiny-unencrypted-9.5",
 			"valid_from": "2017-01-01",
-			"plan_guid": "e264800e-20cb-4bf0-99fc-84bd42681d81",
+			"plan_guid": "5f2eec8a-0cad-4ab9-b81e-d6adade2fd42",
 			"storage_in_mb": 5120,
 			"memory_in_mb": 0,
 			"components": [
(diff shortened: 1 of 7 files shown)


keymon:
I tested this with prod data following the notes in https://github.com/alphagov/paas-billing/pull/49#issuecomment-417677182 and works fine.

keymon:
Thx @henrytk for your comments in #49. The new implementation is simpler, very little changes in code. Most changes were to add tests. It also avoids having to change the semantics of the API. :+1:

henrytk:
I've rebased this on master because the Compose changes (removing its collector) were breaking it on deployment.

PR #53, opened 2018-09-06
WIP: [#159234320] Add consolidation of billable events by month

keymon:
What
----

We want to persist/freeze the billing statement data at the end of each
month, so the bills do not change between months due whatever reason:
 - code changes
 - plan, currency changes
 - new events.

We also want performance querying a bill to improve.

In our implementation we decided to compute and store all the billable
events in a consolidated table, stored per full month.

We implemented in a way that can later be extended to:
 - Allow multiple consolidations to co-exist, so we can compare
   between changes. (e.g. to emit credit notices).
 - Query the history.

Implementation details:

 - We added two tables:
    - `consolidation_history`: Ranges for which we have run
      the consolidation and other data (currently only timestamp).
    - `consoliidation_billable_events`: All the events computed for
      the given consolidation.

 - We use the same query that generate the "realtime" billable events to
   generate the consolidated events, storing the raw json `price` after
   computation.

 - We consolidate FULL months (from 1st to 1st only).

 - Currently we consolidate all the full months from 2017-07-01
   (when we started storing events) till the last month passed the 5th
   of each month.

 - We do not consolidate until the 5th of the next month to ensure we
   have retrieved all service events

 - Currently consolidation only runs once for each full month, and never
   more. We did not implement logic to recompute the consolidation.

 - Currrently To recompute consolidated, you must delete the consolidate data.

 - We extend the `/billable_events` endpoint to split the range of
   dates of the query, and check return the events of is  full month
   that is consolidated

How to review
-----

 - Code review
 - Run tests

Compare with live data
-----------------------

WIP

Who can review
-----

Not @keymon or @richardTowers


keymon:
This is ready to review but I am testing with prod data to compare the generated bills.

I might also do some rebasing.

keymon:
# Verify with prod data 

This has been compared with prod data by import the DB following the instructions in https://github.com/alphagov/paas-billing/pull/14

Start the service locally 

```bash

# Load AWS creds
awssts sts:hectorrivasgandara@gov-paas-dev

# login in CF

PROD_DB_NAME=....

DATABASE_URL=postgres://postgres:@localhost:5432/${PROD_DB_NAME}?sslmode=disable make run-dev DEPLOY_ENV=rafalp

```

Run the comparision

```bash

cf oauth-token > /tmp/prod-token
cf curl  '/v2/organizations?results-per-page=100&q=status:active'| jq -r '.resources[].metadata.guid' > /tmp/prod-orgs.txt

cf login -a https://api.leeporte.dev.cloudpipeline.digital -u admin
cf oauth-token > /tmp/dev-token



TMP_DIR=/tmp/billing-check
mkdir -p $TMP_DIR

get_report() {
    token=$1
    env=$2
    url=$3
    org=$4
    month=$5
    next_month="$(date -d "${month} + 1 month" +%Y-%m-%d)"
    
    mkdir -p "${TMP_DIR}/${month}/"
    curl -qs \
        -H "Authorization: ${token}" \
        "${url}/billable_events?org_guid=$4&range_start=${month}&range_stop=${next_month}" \
        > "${TMP_DIR}/${month}/${org}-${env}.json"
}


sum_report() {
    env=$1
    org=$2
    month=$3
    # The sub is to reduce the precision
    jq  -r 'map(.price.inc_vat | tonumber) | add | tostring | sub("(?<num>[0-9]+\\.[0-9]{4}).*"; .num)' \
        < "${TMP_DIR}/${month}/${org}-${env}.json"
}

gather_reports() {
    start_month="$1"
    orgs="$2"
    for o in ${orgs}; do 
        echo -n "$o: "
        get_report "$(</tmp/dev-token)" dev "${dev_url}" "$o" "$start_month"
        get_report "$(</tmp/prod-token)" prod "${prod_url}" "$o" "$start_month"
        sum_prod=$(sum_report prod $o $start_month)
        sum_dev=$(sum_report dev $o $start_month)
        
        if [ "$sum_prod" == "$sum_dev" ]; then
            echo "ok"
        else 
            echo "differs $sum_prod - $sum_dev = $(echo $sum_prod - $sum_dev  | bc)"
        fi
    done
}

dev_url="http://localhost:8881"
prod_url="https://billing.cloud.service.gov.uk"
orgs="$(</tmp/prod-orgs.txt)"

for m in 2017-{07..12}-01 2018-{01..08}-01; do 
    echo -----------------------
    echo "Testing for $m"
    echo 
    gather_reports ${m} "${orgs}"
done
```

eventio/event_filter.go
```diff
@@ -10,27 +11,118 @@ type EventFilter struct {
...
+		return nil, err
 	}
-	return nil
+	var filters = []EventFilter{}
+
+	nextStartOfMonth := getNextStartOfMonth(start)
+	previousStartOfMonth := getPreviousStartOfMonth(end)
+
+	if end.Before(start) {
+		return []EventFilter{}, nil
```
blairboy362: Is this not an error?

keymon: No, if end < start => no "months" with in it.

Eg. `["9001-01-01" - "2001-01-01"]`

keymon: I think we could refactor all of this in simpler a recursive function like
``` 
get_months(start, end) {
  if end <= start then []
  else 
     next := min(end, truncateMonth(start + "1 month"))
     [(start, next)]::get_months(next, end)
``` 

But I do not want to spend more time on this... tempted though :-D

eventio/event_filter.go
```diff
@@ -10,27 +11,118 @@ type EventFilter struct {
...
+	if err != nil {
+		return *filter, err
+	}
+	stop, err := time.Parse("2006-01-02", filter.RangeStop)
+	if err != nil {
+		return *filter, err
+	}
+
+	return EventFilter{
+		RangeStart: start.Format("2006-01") + "-01",
```
blairboy362: I feel this should use dates and formatting rather than string appends...

```
filterStart = time.Date(start.Year(), start.Month(), 1, 0, 0, 0, 0, start.Location())
...
RangeStart : filterStart.Format("2006-01-02")
```

eventio/event_filter.go
blairboy362: I'm not going to recommend recursion over iteration...better without recursion as a general rule, in my opinion :)

eventserver/server_billable_handler.go
```diff
@@ -26,43 +30,120 @@ func BillableEventsHandler(store eventio.BillableEventReader, uaa auth.Authentic
...
-		if _, err := c.Response().Write([]byte("[\n")); err != nil {
-			return err
-		}
-		c.Response().Flush()
-		next := rows.Next()
-		delim := "\n"
-		for next {
-			b, err := rows.EventJSON()
+		for _, monthFilter := range months {
+			ok, err := consolidatedStore.IsRangeConsolidated(monthFilter)
```
blairboy362: Consider renaming `ok` here. Maybe `isRangeConsolidated`? This isn't a good/bad situation

eventserver/server_billable_handler.go
```diff
@@ -26,43 +30,120 @@ func BillableEventsHandler(store eventio.BillableEventReader, uaa auth.Authentic
...
 			}
-			c.Response().Flush()
+			rowOfRows.RowsCollection = append(rowOfRows.RowsCollection, rows)
+		}
+
+
+		// stream response to client
+		c.Response().Header().Set(echo.HeaderContentType, echo.MIMEApplicationJSONCharsetUTF8)
+		c.Response().WriteHeader(http.StatusOK)
+		WriteRowsAsJson(c.Response(), c.Response(), &rowOfRows)
```
blairboy362: A potential optimisation for the future: this block can be moved around slightly so the writing can be done in a more stream-like way if the reading of the rows into memory starts taking too much memory.

blairboy362: Also: the `error` returned from `WriteRowsAsJson` is being silently swallowed here...

eventio/event_filter.go
keymon: OK, done

keymon: I like it more :-D

eventio/event_filter.go
keymon: done, based on the previous introduced code. I will rebase once you are OK with it.

eventserver/server_billable_handler.go
keymon: done

eventserver/server_billable_handler.go
keymon: > A potential optimisation for the future: I think this block can be moved around slightly so the writing can be done in a more stream-like way if the reading of the rows into memory starts taking too much memory.

But I do not think it is loading them in memory, is it? It opens all the queries to stream from DB at the beginning, but then it reads each one streaming them.

> Also: the error returned from WriteRowsAsJson is being silently swallowed here...

I will fix that

eventserver/server_billable_handler.go
```diff
@@ -26,43 +30,120 @@ func BillableEventsHandler(store eventio.BillableEventReader, uaa auth.Authentic
...
+	for _, r := range r.RowsCollection {
+		err := r.Close()
+		if err != nil {
+			errs = append(errs, err.Error())
+		}
+	}
+	if len(errs) == 0 {
+		return nil
+	}
+	return fmt.Errorf("error calling Close(): %s", strings.Join(errs, " | "))
```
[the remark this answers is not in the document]
keymon: done

keymon:
I rebased this to squash all the latest changes.

eventstore/store_consolidated_billable_events_test.go
```diff
@@ -0,0 +1,359 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		defer db.Close()
+
+		err = db.Schema.Consolidate(eventio.EventFilter{
+			RangeStart: "2001-01-02",
+			RangeStop:  "2001-02-02",
+			OrgGUIDs:   []string{"banana", "pear"},
+		})
+		Expect(err).To(MatchError(
+			"consolidate must be called without an organisations filter (i.e. for all orgs)",
```
blairboy362: Potential future refactor: use a different struct for consolidation filters that doesn't include the possibility of this filter.

eventstore/store_consolidated_billable_events_test.go
```diff
@@ -0,0 +1,359 @@
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Should consolidate events which have not been consolidated yet ", func() {
+		db, err := scenario.Open(cfg)
+		Expect(err).ToNot(HaveOccurred())
+		defer db.Close()
+
+		// Consolidate everything since 2001
+		consolidateSince := "2017-01-01"
```
blairboy362: The comment doesn't match what this line is doing. Please update whichever is correct so they match.

eventstore/store_consolidated_billable_events_test.go
keymon: :+1: TBH it is a quite adhoc thing. 

There is actually no reason to not allow do the consolidation with the OrgGUIDs, but if we do so, the consolidation key should include the list of OrgGUIDs for which it is valid, and that would make it more complicated.

To add a new type, would work but we would have to do some more work to convert types. We could use a compose type or a interface for the intervals to avoid this, but that would silently accept any filter even with orgGUIDs.  

Throwing an error was our way to be explicit saying that it is done for all orgs, and the guids will be ignored.

eventstore/store_consolidated_billable_events_test.go
keymon: damn comments

eventstore/store_consolidated_billable_events.go
```diff
@@ -0,0 +1,319 @@
...
+	"code.cloudfoundry.org/lager"
+	"github.com/alphagov/paas-billing/eventio"
+)
+
+const (
+	DefaultConsolidationStartDate = "2017-07-01"
+)
+
+func (e *EventStore) GetConsolidatedBillableEventRows(filter eventio.EventFilter) (eventio.BillableEventRows, error) {
+	tx, err := e.db.Begin()
```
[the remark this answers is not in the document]
keymon: mmm you are right. 

It is also wrong here: https://github.com/alphagov/paas-billing/blob/66eb9d4add24d3054c8f8712c5f9d543bebf5db2/eventstore/store_billable_events.go#L18-L30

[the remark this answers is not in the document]
keymon: Adding the `tx.Commit()` blows it up

```diff
diff --git a/eventstore/store_billable_events.go b/eventstore/store_billable_events.go
index 75e35b6..6a490c1 100644
--- a/eventstore/store_billable_events.go
+++ b/eventstore/store_billable_events.go
@@ -25,7 +25,7 @@ func (s *EventStore) GetBillableEventRows(filter eventio.EventFilter) (eventio.B
                tx.Rollback()
                return nil, err
        }
-       return rows, nil
+       return rows, tx.Commit()
 }
 
 func (s *EventStore) getBillableEventRows(tx *sql.Tx, filter eventio.EventFilter) (eventio.BillableEven
diff --git a/eventstore/store_consolidated_billable_events.go b/eventstore/store_consolidated_billable_e
index c1480a5..06f65c8 100644
--- a/eventstore/store_consolidated_billable_events.go
+++ b/eventstore/store_consolidated_billable_events.go
@@ -25,7 +25,7 @@ func (e *EventStore) GetConsolidatedBillableEventRows(filter eventio.EventFilter
                tx.Rollback()
                return nil, err
        }
-       return rows, nil
+       return rows, tx.Commit()
 }
````

We get a lot of errors like:

```
      <*errors.errorString | 0xc42032f640>: {
          s: "unexpected command tag SELECT",
      } 
```

Those two methods are readonly, so I guess it makes sense to not create a transaction? :thinking:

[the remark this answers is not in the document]
keymon: OK, now I know what is going on.

The object `BillableEventsRows` has a `Close()` that does the `tx.Rollback()` to avoid leaking transactions. The new and the old method use the same logic:

https://github.com/alphagov/paas-billing/blob/66eb9d4add24d3054c8f8712c5f9d543bebf5db2/eventstore/store_billable_events.go#L93-L98

It is indeed confusing, and I did miss it. Maybe there is a better way to implement this, e.g. by receiving a context or so? Not sure if we should fix this in our PR

eventstore/store_consolidated_billable_events_test.go
blairboy362: Yeah. That's fine. It was just a random thought. Obviously going to be horrid to implement at this point in time and almost certainly not worth the effort.

eventstore/store_consolidated_billable_events.go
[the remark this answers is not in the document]
blairboy362: Where possible we should have the open and corresponding close in the same scope for these things. If it's not a major refactor we should consider it. I wouldn't mind pairing on it if you want.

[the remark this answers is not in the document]
keymon: I think this can be implemented by injecting a `Context` into that rows object that would be closed in the `Close()`, and then starting a go routine waiting for that context to be done to call Rollback. I will give it a go now.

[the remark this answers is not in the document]
keymon: @blairboy362 the idea is this: ef278152bfdb544bae4d46a53e84b576e040d18d

What do you think?

The only doubt I have is that now `rows.Close()` won't return an error. I wonder if it is worth it to have a way to get the error back or not.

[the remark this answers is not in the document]
keymon: Other implementation can be: 9bd7ee45dc76a53e33ae63b1252afb001ac06b06

That simple uses a callback to be called from `billableRows.Close()`. That way we define the `tx.Rollback()` and `rows.Close()` next to the creation of the transaction.

keymon:
@blairboy362 feel free to merge it. 

I will create the PR for paas-cf to get it deployed.

PR #56, opened 2018-11-19
[#161316533] Split the API server and event collector

jpluscplusm:
What
----

This PR splits the API server and event collector into 2 separately-runnable components, in the same `paas-billing` binary.

We're doing this because we need to make the API server blue-green-deployable. This means that it needs to be possible to run it twice, simultaneously, against the same CF.

This is a problem within the current architecture as the collector and the API both run in the same process and we can't allow the collector to run more than once as we're unsure if this might have an effect on the bills produced.

The net result of this is that we need to split this application into 2, the API and the collector, so that we can blue-green-deploy the API portion whilst leaving the deployment of the collector as a simple `cf push` (which results in no more than one instance running).

We achieve this here by adding a second-level noun to the program's invocation and starting only those sections of the program as are required.

To use this: where previously you ran `./path/to/paas-billing`, now to start the collector you should run `./path/to/paas-billing collector` and to start the api/event server you should run `./path/to/paas-billing api`.

We haven't implemented the ability to run both components at once.

How to review
-----

- Code review.

- Smoke-tests: as we began this story, the smoke tests weren't passing on master as they've bitrotted due to not being run anywhere automatically. We've fixed a few bits up, but we ran out of time to get them back to a known-good state. Therefore, unless the reviewer wants to burn some time on getting `master` smoke tests passing, then seeing if this branch makes things worse, we suggest not running them. Possibly `pending-discuss` a story to knock them into shape and into a pipeline somewhere.

- Dev env: we have tested there are no obvious (double-counting; zero-counting) issues in the `jonmtest` env by:
  - starting a large app
  - waiting for over 30 minutes
  - stopping the app
  - waiting for the next 30m cadence of bill generation/update to ocurr
  - observing that the right charge for this app is visible on the current month's bill on paasmin.

Who can review
-----

Anyone except @bandesz and @jpluscplusm

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"database/sql"
+	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -84,24 +72,39 @@ func cfDataCollector(databaseUrl string, logger lager.Logger) error {
 	return nil
 }
 
-func Main(logger lager.Logger) error {
+func Main(ctx context.Context, logger lager.Logger) error {
```
keymon: Nice one this Main context! :+1:

keymon:
I had to add a new commit to revert the manifest format. We might need to update the versions of the clients and plugins in our containers used in concourse

keymon:
Works as announced. I will wait meanwhile it runs for a while in my env, merge it later.

PR #58, opened 2019-01-16
[#163250921] Add larger Redis instance to plans

LeePorte:
What
----

The new plan has been added to the billing config so that we can charge
for it. This is quite a pricey service so we need to ensure we can charge from the start.

How to review
-----

Code review alongside https://github.com/alphagov/paas-cf/pull/1713

Who can review
-----

Not @LeePorte

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,0 +1,980 @@
+{
+	"currency_rates": [
+		{
+			"code": "GBP",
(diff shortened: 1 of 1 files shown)


LeePorte:
I'd missed that we'd move the config. Closing it.

PR #62, opened 2019-02-18
[#163651298] Update process that re-creates the events table

mogds:
What
----

The events table is currently dropped and re-created every 30mins. During this process, the billing page and pricing calculator queries are blocked for several minutes. 

This PR updates the process to create temporary tables that are renamed and run in separate transactions so the queries are unblocked. Also includes a clean up of the SQL files.

How to review
-----

- Code Review
- Deploy to a dev env using this [paas-cf branch](https://github.com/alphagov/paas-cf/pull/1758)

Who can review
-----

Not @mogds @bandesz

### DIFF
--- a/eventstore/sql/create_billable_event_components.sql
+++ b/eventstore/sql/create_billable_event_components.sql
@@ -1,5 +1,6 @@
+DROP FUNCTION IF EXISTS generate_billable_event_components();
 
-CREATE TABLE billable_event_components (
+CREATE TABLE billable_event_components_temp (
 	event_guid uuid NOT NULL,
 	resource_guid uuid NOT NULL,
 	resource_name text NOT NULL,
@@ -16,7 +17,7 @@ CREATE TABLE billable_event_components (
 	memory_in_mb numeric NOT NULL,
 	storage_in_mb numeric NOT NULL,
 	component_name text NOT NULL,
-	component_formula text NOT NULL, 
+	component_formula text NOT NULL,
 	currency_code currency_code NOT NULL,
 	currency_rate numeric NOT NULL,
 	vat_code vat_code NOT NULL,
@@ -26,7 +27,7 @@ CREATE TABLE billable_event_components (
 	CONSTRAINT no_empty_duration CHECK (not isempty(duration))
 );
 
-CREATE OR REPLACE FUNCTION generate_billable_event_components() RETURNS SETOF billable_event_components AS $$
+CREATE OR REPLACE FUNCTION generate_billable_event_components() RETURNS SETOF billable_event_components_temp AS $$
 	with
 	valid_pricing_plans as (
 		select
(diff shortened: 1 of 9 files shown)


eventstore/sql/create_events.sql
```diff
@@ -310,8 +310,17 @@ INSERT INTO events with
...
+CREATE INDEX events_duration_temp_idx ON events_temp using gist (duration);
+CREATE INDEX events_plan_temp_idx ON events_temp (plan_guid);
+DROP TABLE IF EXISTS events;
+ALTER TABLE events_temp RENAME TO events;
+ALTER INDEX events_temp_pkey RENAME TO events_pkey;
+ALTER INDEX events_org_temp_idx RENAME TO events_org_idx;
+ALTER INDEX events_space_temp_idx RENAME TO events_space_idx;
+ALTER INDEX events_resource_temp_idx RENAME TO events_resource_idx;
+ALTER INDEX events_duration_temp_idx RENAME TO events_duration_idx;
+ALTER INDEX events_plan_temp_idx RENAME TO events_plan_idx;
```
richardTowers: Is it necessary to do this? If we just call the indexes the right thing to begin with won't they follow the move?

LeePorte: Sadly you can't as index names are global in Postgres.

richardTowers: Ah, of course.

eventstore/sql/create_events.sql
```diff
@@ -310,8 +310,17 @@ INSERT INTO events with
...
-CREATE INDEX events_resource_idx ON events (resource_guid);
-CREATE INDEX events_duration_idx ON events using gist (duration);
-CREATE INDEX events_plan_idx ON events (plan_guid);
+
+CREATE INDEX events_org_temp_idx ON events_temp (org_guid);
+CREATE INDEX events_space_temp_idx ON events_temp (space_guid);
+CREATE INDEX events_resource_temp_idx ON events_temp (resource_guid);
+CREATE INDEX events_duration_temp_idx ON events_temp using gist (duration);
+CREATE INDEX events_plan_temp_idx ON events_temp (plan_guid);
+DROP TABLE IF EXISTS events;
```
richardTowers: Might help to have a bit of extra whitespace in here.

```suggestion

DROP TABLE IF EXISTS events;

```

eventstore/store.go
```diff
@@ -93,10 +93,23 @@ func (s *EventStore) Init() error {
 			return err
 		}
 	}
-	if err := s.refresh(tx); err != nil {
+	if err := tx.Commit(); err != nil {
 		return err
 	}
-	if err := tx.Commit(); err != nil {
+	// create the ephemeral configuration objects (pricing/plans/etc)
```
richardTowers: Generally we avoid using comments as prolifically as this - the code should speak for itself. I think most of these new comments don't add much value, and may drift out-of-sync with what the code does (at which point they'd be actively harmful).

eventstore/store.go
```diff
@@ -106,67 +119,71 @@ func (s *EventStore) Init() error {
 // Refresh triggers regeneration of the cached normalized view of the event dat and rebuilds the
 // billable components. Ideally you should do this once a day
 func (s *EventStore) Refresh() error {
+	// drop all the projection data
+	if err := s.refreshSQLFile("drop_ephemeral_objects.sql"); err != nil {
+		return err
+	}
```
richardTowers: I wonder if there might be a bit of an issue here now we're splitting up the steps into little transactions.

Previously, we'd drop (for example) the `billable_event_components` table here, then recreate it, all in the same transaction. Now we're dropping it in one transaction (at which point I think the queries will start erroring), and then creating it again later (`create_billable_event_components.sql` further down).

I think you could reproduce this in a dev environment by putting a big sleep in the code here to recreate the situation we have in prod where `create_events.sql` takes several minutes.

I wonder if we might need to create temp tables and then move them for all the "ephemeral" tables, as they're probably all keyed off each other. 🤔

eventstore/sql/create_billable_event_components.sql
```diff
@@ -1,5 +1,7 @@
+DROP FUNCTION IF EXISTS generate_billable_event_components();
```
bandesz: this drop is not needed

eventstore/store.go
```diff
@@ -654,7 +635,7 @@ func checkPlanConsistency(tx *sql.Tx) error {
 		from
 			events
 		where
-			events.plan_guid not in (
+		events.plan_guid not in (
```
bandesz: we should revert this change

eventstore/sql/create_custom_types.sql
```diff
@@ -0,0 +1,17 @@
...
+CREATE TYPE currency_code AS ENUM ('USD', 'GBP', 'EUR');
+EXCEPTION
+    WHEN duplicate_object THEN null;
+END $$;
+
+DO $$ BEGIN
+CREATE TYPE resource_state AS ENUM ('STARTED', 'STOPPED');
+EXCEPTION
+    WHEN duplicate_object THEN null;
+END $$;
```
bandesz: add extra line break

eventstore/store.go
```diff
@@ -677,20 +658,42 @@ func checkPlanConsistency(tx *sql.Tx) error {
...
+			"sqlFile": filename,
+			"elapsed": time.Since(startTime),
+		})
+	}()
+
 	schemaFilename := schemaFile(filename)
 	sql, err := ioutil.ReadFile(schemaFilename)
 	if err != nil {
-		return fmt.Errorf("failed to execute sql file %s: %s", schemaFilename, err)
+		return err
```
bandesz: we should revert this

alext:
This is quite hard to review because the bulk of the changes are in a single commit, and the commit message doesn't currently explain why most of the changes are being made.

Can we update it so that the reasoning behind the changes is captured - this could either be with more detail in the existing commit message, or by splitting it into smaller commits.

bandesz:
@alext I split the changes into smaller commits with meaningful messages.

PR #66, opened 2019-04-24
[#164493738] Billable events are already in GBP

46bit:
What
----

This PR improves the `/billable_events` endpoint (the one whose output is directly what we will bill tenants):

* The prices were already being output in `GBP`, but the `currency_code` field was whatever underlying currency was used by the pricing plan;
* The `currency_rate` didn't make sense to output when the currency has already been converted to `GBP`.

How to review
-----

* Code review;
* 🐝 **Do not merge until https://github.com/alphagov/paas-billing/pull/70 and https://github.com/alphagov/paas-admin/pull/188 have reached production.** 🐝

Who can review
-----

Not @46bit


46bit:
This will require further work to `paas-admin` and `paas-billing` to keep `paas-admin`'s billing statements working. Do not merge until that work is added in.

PR #78, opened 2019-05-09
[#165534387] Switch to Go modules

46bit:
What
----

Switches `paas-billing`'s Go dependency system from `dep` to `go mod`. This simplifies running #77's acceptance tests in Concourse (https://github.com/alphagov/paas-cf/pull/1889) because they will not need to be at `$GOPATH/src/github.com/alphagov/paas-billing`.

How to review
-----

`make bin/paas-billing` and `make test` both succeed for me.

Who can review
-----

Not @46bit.


46bit:
> At the moment we're still planning to vendor the modules (you can do this with `go mod vendor`). This is because otherwise getting the dependencies will be less reliable than the least reliable upstream (e.g. github / code.cloudfoundry.org).
> 
> Once there are proper go mod mirrors we can think about not vendoring dependencies.

Agreed but I've already run `go mod vendor` a few times. The vast numbers of files removed is interesting. I've seen elsewhere that `go mod vendor` only vendors the parts of packages that are actually used, whereas `dep` would vendor the entire package and subpackages, so that might be why.

I'll double-check.

richardTowers:
Oh! I just assumed you'd removed them all because there were so many - sorry!

46bit:
It's OK! My own lack of confidence is why I asked you to review.

I've done some digging and am trying to finish off this change, but it's complex.

https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away describes how to get the `vendor` directory used, using a `go -mod=vendor` command line flag. Building uses `vendor` when that is used.

Ginkgo [added support for that command line flag in its v1.8.0](https://github.com/onsi/ginkgo/issues/559), but errors out if providing that flag to pre-v1.8.0. https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md describes how to pin particular versions of tools, but it seems to be `go generate`-specific 😞

46bit:
In theory this is a great idea, but in practice I want to defer this work until our story about switching to `go mod` everywhere. We'll need to upgrade https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/cf-acceptance-tests to provide a new-enough version of Go, and testing that could snowball. I'll link to this PR on that `go mod` story so it can be picked up then.

PR #81, opened 2019-07-26
[#165537092] Log time elapsed by various database and processing operations

46bit:
What
----

This PR adds logging around the execution time of database queries and calls made to external APIs.

The only database queries whose execution time isn't logged are the Forecast Events–those are a bit awkward to log because it does lots of queries and it's not very important.

How to review
-----

Code review.

Who can review
-----

Not @46bit


46bit:
Non-JSON logs from the API do not appear to be making it into Kibana. For instance, request logs.

PR #82, opened 2019-07-29
[#165537092] Upgrade echo library to log API requests

46bit:
What
----

Newer versions of the echo library do JSON by default, which is something we want. This PR switches dependency management to `go mod` and then upgrades to the newest version of Echo (about 243 commits and 2 years newer.)

We then configure echo to log all requests received.

How to review
-----

* See if the tests go green;
* I suggest code reviewing commit-by-commit. Some of the changes may be hard to understand without reading the commit messages.

Apologies for the large PR–I wanted to avoid having two PRs depending on each other and the multiple-fixing needed during review.

Who can review
-----

Not @46bit.


LeePorte:
The only comment I have is that we have travis using go 1.11.

apiserver/server_logger.go
```diff
@@ -153,7 +154,31 @@ func (l *Logger) SetHeader(_ string) {
...
+// * Always say you wrote all the data.
+// * Never return an error?
+// * Parse the data as JSON without ever modifying the slice.
+func (l *Logger) Write(p []byte) (int, error) {
+	logMessage := map[string]interface{}{}
+	err := json.Unmarshal(p, &logMessage)
+	if err != nil {
+		return 0, err
+	}
+	l.lager.Info("logger-middleware", logMessage)
```
46bit: **Something worth calling out for the review:**

This sets the `@message` field in Kibana for every API request to just be `logger-middleware`, rather than something more informative like `HTTP 200 GET /vat_rates?…`. Maybe we should change that…

LeePorte: It seems a prudent approach to me as the fields are indexable.

PR #83, opened 2019-07-29
[#167483203] Add costs endpoint

mogds:
What
----

- Adds a total costs by plan endpoint so we can monitor the cost
- Adds a `cost_for_duration` column to the `billable_event_components` table

How to review
-----

- Code review

Who can review
-----

Not me

### DIFF
--- a/apiserver/server.go
+++ b/apiserver/server.go
@@ -43,6 +43,7 @@ func New(cfg Config) *echo.Echo {
 	e.GET("/forecast_events", ForecastEventsHandler(cfg.Store))
 	e.GET("/usage_events", UsageEventsHandler(cfg.Store, cfg.Authenticator))
 	e.GET("/billable_events", BillableEventsHandler(cfg.Store, cfg.Store, cfg.Authenticator))
+	e.GET("/totals", TotalCostHandler(cfg.Store))
 
 	e.GET("/", status)
 
--- a/apiserver/server_costs_handler.go
+++ b/apiserver/server_costs_handler.go
@@ -1,0 +1,19 @@
+package apiserver
+
+import (
+	"net/http"
+
+	"github.com/alphagov/paas-billing/eventio"
+	"github.com/labstack/echo"
+)
+
+func TotalCostHandler(store eventio.TotalCostReader) echo.HandlerFunc {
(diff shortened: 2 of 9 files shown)


apiserver/server.go
```diff
@@ -43,6 +43,7 @@ func New(cfg Config) *echo.Echo {
 	e.GET("/forecast_events", ForecastEventsHandler(cfg.Store))
 	e.GET("/usage_events", UsageEventsHandler(cfg.Store, cfg.Authenticator))
 	e.GET("/billable_events", BillableEventsHandler(cfg.Store, cfg.Store, cfg.Authenticator))
+	e.GET("/costs", TotalCostHandler(cfg.Store))
```
richardTowers: I wonder if `/totals`, `/total_costs`, or `/total_costs_by_plan` would be a better name for the endpoint?

apiserver/server_costs_handler.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"net/http"
+
+	"github.com/alphagov/paas-billing/eventio"
+	"github.com/labstack/echo"
+)
+
+func TotalCostHandler(store eventio.TotalCostReader) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		planGUIDSByCost, err := store.GetTotalCost()
```
richardTowers: `planGUIDSByCost` makes me think the return value is a `map[PlanGuid]Cost`, but it's actually a `[]TotalCost`. I think `costTotals` would be a better variable name.

eventstore/store_total_costs.go
```diff
@@ -0,0 +1,40 @@
...
+
+var _ eventio.TotalCostReader = &EventStore{}
+
+func (s *EventStore) GetTotalCost() ([]eventio.TotalCost, error) {
+	tx, err := s.db.Begin()
+	if err != nil {
+		return nil, err
+	}
+
+	defer tx.Rollback()
```
richardTowers: Can we do this without the transaction? I don't think it makes any difference, but the code would be a bit simpler without the unnecessary rollback. (Unnecessary since we're only running one `select`)

PR #86, opened 2019-08-16
[WIP] - Remove compose specific code

richardTowers:
This needs  more thought than I've got time to give it at the moment, but I think we can make some SQL more simple and drop the compose specific tables. Any bills that involve compose will already be consolidated.


richardTowers:
Closing some of my old Draft PRs

PR #87, opened 2019-08-28
Experimental optimization of create_events query

risicle:
Humour me.

This isn't so much meant for merge, but I'd be very interested to know if, run against a database and dataset that I don't have, the following changes make any improvement to the query time of the large `INSERT` statement...

The two changes are independent so should be testable separately if so desired...


risicle:
Another option would be to just make the existing e.g. `(raw_message->>'state')` indexes partial, only where `raw_message->>'space_name' !~ '^(SMOKE|ACC|CATS|PERF)-'`. But I didn't know if those indexes are used elsewhere and this would make them useless for that.

benvand:
Can we WIP this PR if it's experimental?

richardTowers:
The code looks good to me - it just needs someone from the paas team to make time to test it. If it works and there's any kind of performance improvement then I think we'd be happy to merge it.

risicle:
Superseded by #166

PR #89, opened 2019-10-21
[#169035672] Allow a new exchange rate on any day

46bit:
What
----

In https://github.com/alphagov/paas-cf/pull/2110 and https://github.com/alphagov/paas-cf/pull/2111 we tried to change the USD–GBP exchange rate that we use. This failed because of a constraint requiring currency rates to take effect at midnight at the start of a month:

```
new row for relation "currency_rates" violates check constraint "valid_from_start_of_month"
```

This PR removes the requirement that currency rates are only changed at midnight at the start of the month. We will then be able to change currency rates on any day. They still have to take effect at midnight, to make things easier for any future improvements we make to billing.

How to review
-----

* See if the tests go green;
* Code review;
* Deploy to your dev env after deploying https://github.com/alphagov/paas-cf/pull/2111. Restart `paas-billing-collector` a few times and check it doesn't error.

Who can review
-----

Not @46bit.

### DIFF
--- a/eventstore/sql/create_base_objects.sql
+++ b/eventstore/sql/create_base_objects.sql
@@ -153,14 +153,23 @@ CREATE TABLE IF NOT EXISTS currency_rates(
 	rate numeric NOT NULL,
 
 	PRIMARY KEY (code, valid_from),
-	CONSTRAINT rate_must_be_greater_than_zero CHECK (rate > 0),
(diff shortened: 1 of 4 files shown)


chrisfarms:
IIRC only reason this constraint existed was so that the pricing calculator wasn't confusing... this is betterer :+1:

PR #94, opened 2020-12-04
[#175935956] Use Postgres 12 instead of 9.5 for tests

46bit:
What
----

Until now we've used Postgres 9.5 for the billing database in most of our environments. But Postgres 9.5 is being deprecated soon, so we're upgrading to Postgres 12.

This PR changes our tests so that they use Postgres 12. The setup in `travis.yml` is based on https://stackoverflow.com/a/62037133; nothing simpler worked.

How to review
-----

* Code review
* Known caveat: there seems to be some timezone weirdness when running the tests locally. Travis says some timestamps are the same but my local tests disagree. I'm not sure what to do about that, so I'm leaving it for now.

Who can review
-----

Anyone except @46bit

⚠️ You need to sign merges to this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 sudo: false
 
 language: go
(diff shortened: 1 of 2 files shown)


schmie (approved):
LGTM / is needed for a current change

PR #95, opened 2020-12-21
Increase DefaultRefreshTimeout

mogds:
What
----

The refresh function in prod london is timing out and currently fails when
running create_billable_event_components.sql. We're in the process of
reviewing billing so this increases the timeout to resolve the issue for now.

It also increases the processor schedule from 30 minutes to 1 hour.

How to review
-----

Code review

Who can review
-----

Not me


AP-Hunt (approved):
I don't know of any other places that need amending

PR #97, opened 2021-02-22
Update formulae for tenants upgrading their RDS instances

poveyd:
What
----

Update formulae for tenants upgrading their RDS instances. The new formulae are only updated for the month(s) in which upgrades occur. This is because tenants were being overcharged for the months in which they upgraded their RDS databases.

Formulae are being changed from:

`'($storage_in_mb/1024) * ceil($time_in_seconds/2678401) * 0.127'`
to
`'($storage_in_mb/1024) * 0.127 * ceil($time_in_seconds)/2678400'`

and from:
        
`'($storage_in_mb/1024) * ceil($time_in_seconds/2678401) * 0.253'`
to
`'($storage_in_mb/1024) * 0.253 * ceil($time_in_seconds)/2678400'`

Position of the brackets is being changed to round charge up to nearest second rather than to nearest month.

How to review
-----

The code was run before and after the change, extracting the contents of billable_event_components after each run. Differences between each set of results was reviewed.

Who can review
-----

Someone with knowledge of Paas billing.

### DIFF
--- a/eventstore/sql/create_billable_event_components.sql
+++ b/eventstore/sql/create_billable_event_components.sql
@@ -28,36 +28,57 @@ CREATE TABLE billable_event_components_temp (
 	CONSTRAINT no_empty_duration CHECK (not isempty(duration))
 );
 
-CREATE OR REPLACE FUNCTION generate_billable_event_components() RETURNS SETOF billable_event_components_temp AS $$
-	with
-	valid_pricing_plans as (
-		select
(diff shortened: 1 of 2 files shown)


AP-Hunt (approved):
This is a little difficult to review. I don't see anything glaringly wrong in the code, and I think I understand the formula change.

However, this could definitely do with a second reviewer before merging.

PR #101, opened 2021-05-06
Add an owner column into the API

paroxp:
What
----

We'd like to label our metrics with the owner. Adding it into the
Database when the orgs are being gathered is the easiest option at the
moment...

This can only be accommodated with the V3 API, so making a custom call.

The client has been regenerated with counterfeiter.

```
counterfeiter -o fakes/fake_cf_data_client.go cfstore/cfstore_client.go CFDataClient
```

How to review
-----

- Code review
- Run tests
- Run the app in safe environment and expect it to fill in the data properly
- See if my SQL changes are good or over thought...

### DIFF
--- a/cfstore/cfstore.go
+++ b/cfstore/cfstore.go
@@ -241,19 +241,22 @@ func (s *Store) collectOrgs(tx *sql.Tx) error {
 				name,
 				created_at,
 				updated_at,
-				quota_definition_guid
+				quota_definition_guid,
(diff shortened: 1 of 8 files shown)


schmie (approved):
- Tests finally pass (merged #94 for tests to use postgres-12 in the end).
- Changes look good to me - `owner` field is filled correctly in my dev and the collector works fine.

PR #105, opened 2021-05-24
[#176541815] Create a performance data exporter

AP-Hunt:
What
----
In story 176541358 (Pivotal), we build scripts to pull out performance data from the billing system. In this story (176541815) we chose to automate writing it to Google Sheets, so that it can be viewed in Google Data Studio.

The code itself is simple, and we (@mogds  and I) believe it probably doesn't require any tests right now:
1. Run SQL scripts
2. Upload to Google Sheets

There remains an open question for the team: how should we handle the exporters in dev envs? We don't want dev envs to have to rely on an arbitrary Google Sheet to function, nor do we want to complicate the pipelines ([paas-cf pr](https://github.com/alphagov/paas-cf/pull/2665)) by trying to (not) deploy the CF app depending on the environment. We propose 3 options:

1. Introduce a flag to write the data to stdout instead of Google Sheets (could require introducing some tests, depending on complexity/teams comfort level)
2. Give it an empty set of Google Sheets credentials and let it error silently
3. Give every environment the same sheet to write to

How to review
-----
1. Code review (ask for a pair if you want one)
2. See it deployed in `dev01`
3. Look at the `paas-credentials` PR (linked in Pivotal)
4. Look at the [`paas-cf` PR](https://github.com/alphagov/paas-cf/pull/2665)

---


LeePorte:
I would lean towards `Introduce a flag to write the data to stdout instead of Google Sheets (could require introducing some tests, depending on complexity/teams comfort level)`. This enables quick feedback and reduces complexity when setting up a dev env from scratch.

PR #106, opened 2021-05-25
#177457361 Set a default value for `owner` in `orgs` table

schmie:
# What
Not all orgs in our production CF database seem to have the owner metadata set (e.g. tests). This breaks the (re-)generation of the orgs table in the `billing-db`.

Setting a default `Owner not set` value of `owner` in the `orgs` table so that the generation of the table does not fail.

How to review
-----
- Code review
- Check SQL to be working in a dev env (or staging for that matter). The `owner` column may need to be dropped for the logic to work properly.

Who can review
-----

not @schmie 

---


paroxp (approved):
Seems legit!

PR #108, opened 2021-06-07
Add script generating corrections SQL

paroxp:
What
----

We believe the current solution, doesn't synchronise the data with the
CF API. Instead of adding that functionality directly to the software,
we're creating a script that is simply generating bunch of update
clasures, making sure the data is consistent.

This can be run ad-hoc and any time, but the outputs, will need to be
manually applied to the postgres DB.

How to review
-----

- Setup local postgres DB
- Create the table and fill it with some data
- Run the script
- Apply the generated patch

---

### DIFF
--- a/scripts/synchronize/.gitignore
+++ b/scripts/synchronize/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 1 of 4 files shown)


schmie (approved):
Tested with prod dataset :+1: 
Could include semicolons, but not a blocker

paroxp:
I was curious about the semis... I wasn't sure if postgres would spit it out or not. I'm happy to add them?

schmie (approved):
:) thanks

PR #123, opened 2021-10-27
Gherkin test

whpearson:
What
----

Gherkin test framework, the actual tests will live in paas-cf, see https://github.com/alphagov/paas-cf/pull/2766

How to review
-----

Deploy to a dev environment with the above PR deployed for paas-cf

Who can review
-----

---


gherkin/billing_charges_test.go
```diff
@@ -0,0 +1,585 @@
...
+	defaultSpaceName = "test-space-name"
+
+	ctx.BeforeSuite(func() {
+		fmt.Println("Connecting to the database")
+		conn := "user=postgres dbname=postgres host=localhost sslmode=disable"
+		db, err = sql.Open("postgres", conn)
+		if err != nil {
+			panic(err)
+		}
+		// defer db.Close()
```
cadmiumcat, line 63: Unless we actually want to db.Close()?
```suggestion
```

gherkin/billing_charges_test.go
```diff
@@ -0,0 +1,585 @@
...
+	tableList := strings.Split(tables, ",")
+
+	for i := 0; i < len(tableList); i++ {
+		sql := fmt.Sprintf("DELETE FROM %s;", tableList[i])
+		// fmt.Printf("Running '%s'\n", sql)
+		rows, err := db.Query(sql)
+		if err != nil {
+			panic(err)
+		}
+		_ = rows
```
cadmiumcat, line 143:
```suggestion
		_, err := db.Query(sql)
		if err != nil {
			panic(err)
		}
```

gherkin/billing_charges_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+	// Add data to the following tables:
+	// vat_rates, currency_rates, pricing_plans, pricing_plan_components.
+
+	planConfig, err := eventstore.LoadConfig("../config.json")
+	if err != nil {
+	  panic(err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 10 * time.Minute )
```
cadmiumcat, line 162: I believe unless we're planning to use a different context, the background one makes more sense
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute )
```

gherkin/billing_charges_test.go
```diff
@@ -0,0 +1,585 @@
...
+			}
+			sql := string(content)
+			rows, err := db.Query(sql)
+			if err != nil {
+				panic(err)
+			}
+			_ = rows
+		}
+
+		// defer db.Close()
```
cadmiumcat, line 100:
```suggestion
```

cadmiumcat (commented):
Tests are passing locally 🥳. I'm currently fixing a dev env to get it running there 

For now I've left a few comments. There are a few more things but I didn't want to swamp you with `suggestions`. Here are some things I think we should deal with: 
- handling errors. We're using `panic` in a few places where we don't want things to crash, we could instead handle the errors.
- removing commented code
- renaming variables that use `_` instead of camelCase

I'm happy to make the changes for the last 2 at least if you'd like? let me know 🙌

gherkin/billing_charges_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+		for i, table := range tables {
+			_ = i
+			table = pathToSqlDefinitions + table
+			fmt.Printf("Creating tables and other database objects in the file: %s...\n", table)
+			content, err := ioutil.ReadFile(table)
+			if err != nil {
+				panic(err)
+			}
+			sql := string(content)
```
cadmiumcat, line 92: This variable shares a name with an imported package - probably best to avoid

whpearson:
All your changes look good @cadmiumcat

cadmiumcat (approved):
Looks good for now 👍

PR #127, opened 2021-12-13
Dp update resources performance and rerun changes

poveyd:
What
----

Changed update_resources and associated event_tables so that events are only processed once when updating the resources table. Before, update_resources used the old billing code to go through all events and build up a story of changes before updating resources. Now, update resources only goes back to the earliest unprocessed event and works forward from that point when building up the story of changes. Anything in the resources table before that point is left unchanged.

How to review
-----

There are no automated tests covering update_resources.

Tests done:

1. Run `update_resources` against to populate `resources` table from empty. Confirmed it produced same results as old version of `update_resources`

2. Run `update_resources` having first set `processed = false` for old event from 6 months ago. Confirmed `resources` table unchanged by this

3. Added new 'TASK_STARTED' event into `app_usage_events` and run `update_resources`. Confirmed new `resources` row created and no other entries in resources table impacted.

Who can review
-----

Not @poveyd
---


whpearson:
It looks like we are not creating the process_lock entry by calling billing-db/static_data/process_lock.sql anywhere at the moment in the go code, or did I miss it?

whpearson:
Also we need adding something to create the table, billing-db/tables/process_lock.sql in eventstore/store.go.

Although maybe we need that after we have added in things from the other PR.

PR #128, opened 2021-12-13
New-billing-integration part 1

whpearson:
What
----
This is part 1 of the new billing integration work. This is the stuff that should not impact current billing at all. It is mainly changing the database tables and stored procedures not currently used by billing with a few ancillary data structures. 

How to review
-----
Deploy to  dev env?

Who can review
-----
Not @whpearson 
---

### DIFF
--- a/billing-db/tables/resources.sql
+++ b/billing-db/tables/resources.sql
@@ -19,5 +19,5 @@ CREATE TABLE IF NOT EXISTS resources
 	last_updated TIMESTAMPTZ NOT NULL -- When this row was last updated
 );
 
-CREATE INDEX CONCURRENTLY resources_i1 ON resources (valid_from, valid_to);
-CREATE INDEX CONCURRENTLY resources_i2 ON resources (org_name, valid_from, valid_to);
+CREATE INDEX IF NOT EXISTS resources_i1 ON resources (valid_from, valid_to);
+CREATE INDEX IF NOT EXISTS resources_i2 ON resources (org_name, valid_from, valid_to);
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 language: go
 
 go:
-  - 1.13
+  - 1.16
 
 addons:
   apt:
--- a/billing-db/sprocs/calculate_bill.sql
+++ b/billing-db/sprocs/calculate_bill.sql
@@ -51,8 +51,8 @@ CREATE TEMPORARY TABLE billable_by_component
     is_processed BOOLEAN NULL
 );
 
-CREATE INDEX CONCURRENTLY IF NOT EXISTS billable_by_component_i1 ON billable_by_component (generic_formula, storage_in_mb, memory_in_mb, number_of_nodes, external_price);
-CREATE INDEX CONCURRENTLY IF NOT EXISTS billable_by_component_i2 ON billable_by_component (generic_formula);
+CREATE INDEX IF NOT EXISTS billable_by_component_i1 ON billable_by_component (generic_formula, storage_in_mb, memory_in_mb, number_of_nodes, external_price);
+CREATE INDEX IF NOT EXISTS billable_by_component_i2 ON billable_by_component (generic_formula);
 
 -- For the billing calculator, we can easily create the billable_by_component table and populate it with what the user wants to get the prices for, then call the calculate_bill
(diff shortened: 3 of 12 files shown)


billing-db/sprocs/get_tenant_bill.sql
```diff
@@ -20,7 +20,7 @@ CREATE TEMPORARY TABLE billable_resources
 );
 
 -- The billable_by_component table needs creating before running this stored function. This is so we can preserve the contents of this table for audit/debug purposes.
-CREATE TEMPORARY TABLE billable_by_component
+CREATE TABLE IF NOT EXISTS billable_by_component
```
poveyd, line 23: The billable_by_component and billable_resources tables need to be created as temporary tables (part of the new billing design). This code won't work otherwise - bills will get incorrectly calculated if more than one thread is running (e.g. more than one tenant bill being calculated at once or if later the billing calculator is being used).

billing-db/sprocs/get_tenant_bill.sql
```diff
@@ -51,8 +51,8 @@ CREATE TEMPORARY TABLE billable_by_component
     is_processed BOOLEAN NULL
 );
 
-CREATE INDEX CONCURRENTLY IF NOT EXISTS billable_by_component_i1 ON billable_by_component (generic_formula, storage_in_mb, memory_in_mb, number_of_nodes, external_price);
-CREATE INDEX CONCURRENTLY IF NOT EXISTS billable_by_component_i2 ON billable_by_component (generic_formula);
+CREATE INDEX IF NOT EXISTS billable_by_component_i1 ON billable_by_component (generic_formula, storage_in_mb, memory_in_mb, number_of_nodes, external_price);
```
poveyd, line 54: It's ok here just to have `CREATE INDEX billable_by_component_i1 ON billable_by_component ...` The `IF NOT EXISTS` is not needed

billing-db/sprocs/get_tenant_bill.sql
whpearson: Thanks, I'll fix that. I think it was just changed for debugging purposes.

billing-db/tables/resources.sql
```diff
@@ -19,5 +19,5 @@ CREATE TABLE IF NOT EXISTS resources
 	last_updated TIMESTAMPTZ NOT NULL -- When this row was last updated
 );
 
-CREATE INDEX CONCURRENTLY resources_i1 ON resources (valid_from, valid_to);
-CREATE INDEX CONCURRENTLY resources_i2 ON resources (org_name, valid_from, valid_to);
+CREATE INDEX IF NOT EXISTS resources_i1 ON resources (valid_from, valid_to);
```
poveyd, line 22: You could have `CREATE INDEX CONCURRENTLY IF NOT EXISTS resources_i1...` here, just in case this is being run against an already-existing instance of `resources`

billing-db/sprocs/get_tenant_bill.sql
```diff
@@ -189,3 +193,278 @@ BEGIN
...
+            r.memory_in_mb,
+            r.number_of_nodes
+    FROM  resources r
+    WHERE r.valid_from < _from_date
+    AND   r.valid_to > _to_date;
+
+    RETURN QUERY
+    SELECT * FROM calculate_bill();
+END
+$$;
```
poveyd, line 470: It's often good style to have each database object, with the exception of simple dependent objects such as indices, constraints, etc., defined in their own file. Not so important with Postgres but can trip you up and cause problems when using other databases

poveyd (changes requested):
The important thing that needs changing is to revert `billable_by_component` and `billable_resources` to being temporary tables, otherwise you will get conflicts between multiple runs of `get_tenant_bill*` with undefined results - tenant bills will be wrong

billing-db/tables/resources.sql
whpearson: Looking back at the commit history of the new-billing-integration it looks like the concurrently key words were removed due to them not being allowed during a transaction. @schmie can you give more context?

schmie: The golang execution within a transaction crashes at CONCURRENTLY.

poveyd (approved):
Looks good

schmie (approved):
:+1: Deploys fine in my dev env, no interaction with components of current billing added, yet.

PR #129, opened 2021-12-16
Dp update resources performance and rerun changes

poveyd:
What
----

The following changes were made:
    
- Changes made to make `update_resources` run more quickly - each event in `app_usage_events`, `service_usage_events`, and `compose_audit_events` is now only processed once (unless a historical event is added). Run time is reduced from about 2 hours to about 4 minutes as a result of this change (based on test involving adding a historical event). A new field has been created in the `app_usage_events`, `service_usage_events`, and `compose_audit_events tables` to identify events that have not yet been processed by `update_resources`. This new field (`processed`) is set to false for new events but is changed to true once the event has been processed by `update_resources`
    
- Change made to ensure only one instance of `update_resources` runs at a given time. If more than one instance runs at the same time then the `resources` table will have incorrect data. This uses the new `process_lock `table.
    
- Some reformatting to tidy up code

How to review
-----

There are no automated tests covering `update_resources`.

Tests done:

1. Run `update_resources` against to populate `resources` table from empty. Conﬁrmed it produced the same results as the old version of `update_resources`

2. Run `update_resources` having ﬁrst set `processed = false` for old event from 6 months ago. Conﬁrmed `resources` table was unchanged by this

3. Added new 'TASK_STARTED' event into `app_usage_events` and run `update_resources`. Conﬁrmed new resources row created and no other entries in `resources` table impacted.

Who can review
-----

Not @poveyd

---

### DIFF
--- a/billing-db/sprocs/update_resources.sql
+++ b/billing-db/sprocs/update_resources.sql
@@ -1,440 +1,538 @@
 -- Do not run too far back in time or will take too long to update resources during which time users won't be able to generate billing reports
-CREATE OR REPLACE FUNCTION update_resources
-(
-    _from_date TIMESTAMPTZ
-)
+CREATE OR REPLACE FUNCTION update_resources()
 RETURNS TABLE
 (
-	num_rows_added INT
+    num_rows_added INT
 )
 LANGUAGE plpgsql AS $$
+DECLARE _from_date TIMESTAMPTZ;
 DECLARE _run_date TIMESTAMPTZ := NOW();
 DECLARE _num_rows_added INT;
+DECLARE _max_compose_event_processed INT;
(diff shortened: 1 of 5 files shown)


whpearson:
It looks like there has been a reversion in this PR, it looks like app_usage_event tables do not have the processed flag added as was done in a previous version of the PR.

whpearson:
With @schmie 's help I connected a locally running dev db to a dev cf instance. In this environment I verified that the PR sets up the required tables and columns, creates resources from the events coming from the environment and also update resources of existing services based on new information.

whpearson (approved):
Looks good

schmie (approved):
:+1: LGTM

PR #133, opened 2022-03-02
Change generateMissingPlans to not look for Epoch

whpearson:
What
----
It is no longer the case that pricing plans have to cover the whole epoch, this was causing problems with the pricing calculator.

Instead make sure that there is a pricing plan that exists before the event in question.

How to review
-----

Deploy to a dev environment and make sure that 'app app' events are not created and that the billing collector can run.

Who can review
-----
Not @whpearson 

---


schmie (approved):
:+1: Tested in dev01/checked against dev03 works as intended.

PR #134, opened 2022-03-18
Stop looking for pricing plan components from the beginning of time

whpearson:
What
----
We now have components that start after the beginning of time EPOCH.

This however confuses the generateMissingPlans function as it still
looks for things that start at EPOCH. So it the tries to
generate new pricing components, with epoch as the starting time and
fails as it does not match the pricing plan (the logic we fixed in this
commit https://github.com/alphagov/paas-billing/pull/133/commits).

This fixes the logic in the same way to make sure we generate a pricing plan
component only if we need one for the current event, that is there is
not a component valid before the start of the event.

How to review
-----
Code review

Who can review
-----

Not @whpearson 

---

### DIFF
--- a/eventstore/store.go
+++ b/eventstore/store.go
@@ -720,7 +720,7 @@ func (s *EventStore) generateMissingPlans(tx *sql.Tx) error {
 				select distinct plan_guid
 				from pricing_plan_components ppc
 				where ppc.plan_guid = events.plan_guid
-				and valid_from = 'epoch'::timestamptz
+				and valid_from < lower(events.duration)
 			)
 		)`,
 	); err != nil {


whpearson:
The error seen in the logs was ```   2022-03-18T15:06:31.41+0000 [APP/PROC/WEB/0] OUT {"timestamp":"1647615991.411420822","source":"paas-billing","message":"paas-billing.exit-error","log_level":2,"data":{"deployment":"prod-lon","error":"generate-service-plan-component: insert or update on table \"pricing_plan_components\" violates foreign key constraint \"pricing_plan_components_plan_guid_valid_from_fkey\": Key (plan_guid, valid_from)=(c07ac6bb-8b26-4a08-b8db-9618d9f5f852, 1970-01-01 00:00:00+00) is not present in table \"pricing_plans\"."}}
```
Plan_guid is c07ac6bb-8b26-4a08-b8db-9618d9f5f852 and is postgres tiny-unencrypted-10 high-iops and was valid from 2020-02-01

schmie (approved):
:+1:

schmie (approved):

schmie:
Not needed anymore with new billing changes reverted

PR #139, opened 2022-04-05
Remove tests that rely on refreshing db and compare string not arrays

whpearson:
What
----
We no longer do the required work when refreshing the DBs for old
billing. This removes the tests that relied upon them.

I also changed some tests that did array comparison to unmarshalling and
then comparing strings, as the two arrays were not the same on the byte
level.

How to review
-----

Validate testing methodology and that no useful tests have been deleted. The additions are in eventstore/store_test.go

Who can review
-----

---


eventstore/store_vat_rates_test.go
```diff
@@ -1,103 +0,0 @@
...
-		}
-
-		env, err := testenv.Open(cfg)
-		Expect(err).ToNot(HaveOccurred())
-		defer env.Close()
-		store := env.Schema
-
-		Expect(store.Refresh()).To(Succeed())
-
-		vatRates, err := store.GetVATRates(eventio.TimeRangeFilter{
```
whpearson, line 95: Adjust to look elsewhere into resources?

eventstore/store_usage_events_test.go
```diff
@@ -1,1295 +0,0 @@
...
-var _ = Describe("GetUsageEvents", func() {
-
-	var (
-		cfg eventstore.Config
-	)
-
-	BeforeEach(func() {
-		cfg = testenv.BasicConfig
-	})
-
```
whpearson, line 23: Look at tests for config -> eventstore we have tests for db insert and requests but not code level stuff currently

PR #145, opened 2022-04-19
cf billing reconcilliation tool

whpearson:
What
----

This tool is to verify that cloud foundries view of the world is a lot
like billings. This is only done for services initially.

It looks at cloud foundry from 4 hours ago, so may miss something, but
this is to take into consideration the time taken for billing to update.

Once it has got a list of services from each API it makes sets of them
using maps.

It then checks the intersection of the maps and the makes sure there are
not too many things in billing and not in cf (which happens for
transient services)

How to review
-----

Follow the readme to run the test

Who can review
-----

Not @whpearson 

---


risicle:
I realize your hands were tied here with the interfaces available to you, but.. seeing what amounts to a single sql query written in golang is like https://giphy.com/gifs/head-pain-YpmVBNubONoqs

reconcilliation-test/cf_billing_reconcilliation.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	var serviceInstances listV3ServiceInstancesResponse
+
+	err = json.Unmarshal([]byte(body), &serviceInstances)
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	for i := 0; i < len(serviceInstances.Resources); i++ {
```
AP-Hunt, line 158: Why not `for _, serviceInstance := range serviceInstances` here?

reconcilliation-test/cf_billing_reconcilliation.go
```diff
@@ -0,0 +1,197 @@
...
+	//Do the billing api request
+	billingAPIURL, err := url.Parse(os.Getenv("BILLING_API_URL"))
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	billingAPIURL.Path = "/billable_events"
+
+	q := billingAPIURL.Query()
+	q.Set("range_start", time.Now().AddDate(0, 0, -2).Format("2006-01-02"))
```
risicle, line 73: What's the relevance of the two-days-ago period start?

reconcilliation-test/cf_billing_reconcilliation.go
```diff
@@ -0,0 +1,197 @@
...
+			cf_service_instance_guids[serviceInstances.Resources[i].Guid] = true
+
+		}
+	}
+
+	fmt.Println("The number of service instances in cf: ", len(cf_service_instance_guids))
+	fmt.Println("The number of service instances in paas-billing: ", len(billing_service_instance_guids))
+
+	//Find the intersection between the two sets of service guids and also the sets of guids in one but not the other
+	set_intersection := get_boolean_results(cf_service_instance_guids, billing_service_instance_guids)
```
risicle, line 174: Surely this variable should be called `boolean_results`?

reconcilliation-test/README.md
```diff
@@ -0,0 +1,20 @@
+
+What
+===
+
+This is a small script that checks whether the data in the cloud foundry database is similar to the data in the billing database about what services exist.
```
risicle, line 5: So, in english, what this does is check that all currently existing service instances older than 4h have billing events in the last 2d, and that there aren't *too many* service instances that have billing events in the last 2 days that don't seem to currently exist? If correct, this might be nice to include here.

risicle:
Think this basically looks right, but did you want to discuss how to make this fit into the pipeline?

reconcilliation-test/cf_billing_reconcilliation.go
whpearson: Fair point I'll update

reconcilliation-test/cf_billing_reconcilliation.go
whpearson: So we get some data but don't go to the beginning of time. @schmie picked that date, so I can't speak to the actual date.

whpearson:
> Think this basically looks right, but did you want to discuss how to make this fit into the pipeline?

Yeah, I was thinking a smoke test that ran against $DEPLOY_ENV every 4 hours?

schmie (approved):
Might need some refinement of filters, but is benign enough to deploy for now

PR #161, opened 2022-11-09
eventstore: add proper migration support using "migrate" package

risicle:
What
----

App was previously "making do" without proper database migration handling & tracking by using schema-creation sql which was able to cope with the presence of existing objects. This only works to a basic degree and can get hairy if we start using dumps/restores of existing databases that may have had different versions of the pseudo-migrations applied to them.

Depends on https://github.com/alphagov/paas-cf/pull/3026

How to review
-----

Who can review
-----

---


eventstore/migrations/0040_create_base_objects.up.sql
```diff
@@ -1,3 +1,8 @@
+-- "migration" written before we had proper migration handling, hence the
+-- various attempts at mitigating previously existing objects
+
+BEGIN;
+
```
corlettb, line 5: Should we get rid of the DROP TABLE's as we now delete the contents of the table on startup?

eventstore/migrations/0010_create_custom_types.up.sql
```diff
@@ -1,3 +1,8 @@
+-- "migration" written before we had proper migration handling, hence the
```
corlettb, line 1: Maybe we should add a comment in these files to say something like:

'NEVER update these files once they have been merged to main. If you need to make changes create a new change file'

Sadly I've seen far too many developers decide to try and change these files after the fact and create a mess.

risicle: :scream: really?

corlettb: I've seen it happen about 5-10 times throughout my carrier sadly.

risicle: Your carrier?

Have put a banner. Though it actually reminds me that something I _do_ sometimes do is add comments to old migrations when e.g. function definitions have been superseded in later migrations.

corlettb (approved):
Looks good after a quick test

schmie (approved):
LGTM 👍

PR #162, opened 2022-11-14
make acceptance target: split into acceptance and integration targets

risicle:
What
----
These are two separate things that perform two different tasks and have different sets of external dependencies. Splitting them allows us to run the `integration` tests *before* deployment and possibly avoid a bad deploy.

The `acceptance` tests poke the deployed billing api and can be used _after_ deployment to check it's gone out ok.

How to review
-----

A little tricky, you can probably see deployments in `dev03` that I've done using https://github.com/alphagov/paas-cf/pull/3026. That PR needs to be merged _after_ this one, but in the intervening time between this one and that one being merged, only half the tests will be running on paas-billing deploys, so merging them in fairly quick succession would be optimal.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,12 +62,15 @@ smoke:
 	echo "smoke test enabled against ${BILLING_API_ADDRESS}"
 	go run github.com/onsi/ginkgo/v2/ginkgo  -focus=".*from api" -r acceptance_tests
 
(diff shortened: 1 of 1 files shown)


corlettb (approved):

risicle:
Urgh forgot this needed signed merges too

PR #166, opened 2022-11-30
eventstore create_events: use custom aggregators to significantly accelerate

risicle:
https://www.pivotaltracker.com/story/show/183803351

What
----

Broadly, this introduces two new custom aggregators and uses them to significantly reduce the query time for the `create_events.sql` operation. `last_agg` can be used to replace some array-based hacks used for populating missing values from previous events in window functions. This reduces the query time from ~12h to ~7m.

`anydistinct` is used to remove redundant rows from the `valid_service_plans`, `valid_services`, `valid_orgs` and `valid_spaces` common table expressions which we later outer-join against. This halves the remaining query time to ~3 minutes.

The entire "refresh" procedure still takes ~30mins but this is now almost entirely the other queries performed as part of that, which I haven't looked at (but could do if needs be).

Please see the individual commits for more detail on each change.

How to review
-----

Look at prod-lon's "billing-experimental" deployment, which should be operating correctly, with approximately 30 min refresh operations.

I have verified that this query produces identical results to the previous one when run against production data. I can show this upon request.

Who can review
-----

Perhaps people want to wait until after the explainer session before reviewing this :shrug: 

---

### DIFF
--- a/eventstore/migrations/0120_last_agg.up.sql
+++ b/eventstore/migrations/0120_last_agg.up.sql
@@ -1,0 +1,22 @@
+-- **do not alter - add new migrations instead**
+
+BEGIN;
+
+
+--
+-- add basic last_agg aggregation function, which simply returns the value most
+-- recently presented to it. powerful when used with aggregation FILTER clause.
+--
+
+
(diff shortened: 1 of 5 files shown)


cadmiumcat (approved):
This is great. Thanks for explaining it so thoroughly
👏

whpearson (approved):

PR #168, opened 2022-12-29
collector initialization: move initial `regenerateEvents()` call to start of event processor loop

risicle:
What
----

https://www.pivotaltracker.com/story/show/184071069

See commit messages for more details.

Running `regenerateEvents()` as part of `EventStore.Init()` means the collector process can't complete initialization of all its' worker loops until that has finished. This can become a problem if `regenerateEvents()` takes a very long time or is likely to fail, in which case the collection worker-loops may never get started at all - which is bad because they gather data which we _may_ not get another chance to.

Instead, ensure the asynchronous event processor worker runs `regenerateEvents()` immediately once it is started, which, as it happens is very (very) shortly after it would have been started in the first place.

How to review
-----

Deploy to a dev env, check the acceptance & integration tests continue to pass.

---

### DIFF
--- a/eventstore/store.go
+++ b/eventstore/store.go
@@ -125,9 +125,6 @@ func (s *EventStore) Init() error {
 		return err
 	}
 
-	if err := s.regenerateEvents(); err != nil {
-		return err
-	}
 	s.logger.Info("initialized")
 	return nil
 }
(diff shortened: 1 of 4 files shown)


risicle:
Closed in favour of  #169

PR #179, opened 2023-06-21
Restrict all-org usage & billable events requests to admins

risicle:
What
----

https://www.pivotaltracker.com/story/show/185167233

The original aim of this was to enforce permissions boundaries in `/usage_events` (and `/billable_events`), but the scope had to expand to put in place some tests that would actually _cover_ this. And I wanted to add _end-to-end_ tests because it was precisely the piecemeal, heavily-mocked unit test approach that allowed this to slip through in the first place.

So this expands the "acceptance tests" (run by our `create-cloudfoundry` pipeline following a live deployment) to also test requests on behalf of "real" admin _and_ non-admin users. The idea being that a failure in staging will prevent deployment of broken code to prod.

Writing these tests also uncovered a few problems with the `/billable_events` handler, due to a previous attempt at saving app memory, though it was at the cost of db server memory and connections. This allowed a single request for a large date range to exhaust the database's connections. See commit comment for more detail.

I removed that mechanism because running out of app memory is far more manageable than running out of db memory & connections. It does however expose the problem that the design of both the `/usage_events` and `/billable_events` endpoints, not having any paging (or even result-limiting) mechanisms, imply to clients they can do the impossible - return an infinite number of results without issue. The current half-solution is to use streaming http responses, but this has the downside that it breaks proper error communication - once the `200` status code has been sent at the beginning of the response, there's no going back on that and any errors encountered later on in the request processing will manifest to the client as either an in-body error message, truncated response or TCP-level failure.

My suggestion would be to either add a basic result-limiting mechanism (hard to shoehorn in to an existing REST interface, and made more complicated by our existing "streaming response" mechanism) or accumulating the response in a gzipped buffer before returning it (echo's "gzip" middleware will likely have the same problem as streaming responses, only starting to compress after the status line is sent).

So instead of making this PR any bigger I left the latter problem unsolved, instead limiting the size of requests used in my acceptance tests.

How to review
-----

---

### DIFF
--- a/apiserver/server_billable_handler.go
+++ b/apiserver/server_billable_handler.go
@@ -2,12 +2,7 @@ package apiserver
 
 import (
 	"context"
-	"fmt"
 	"net/http"
-
-	"io"
-
-	"strings"
 
 	"github.com/alphagov/paas-billing/apiserver/auth"
 	"github.com/alphagov/paas-billing/eventio"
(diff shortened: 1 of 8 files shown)


apiserver/server_billable_handler.go
```diff
@@ -35,118 +42,66 @@ func BillableEventsHandler(store eventio.BillableEventReader, consolidatedStore
...
-	if _, err := writer.Write([]byte("[\n")); err != nil {
-		return err
-	}
-	flusher.Flush()
-	next := rows.Next()
-	delim := "\n"
-	for next {
-		b, err := rows.EventJSON()
-		if err != nil {
+		if _, err := c.Response().Write([]byte("\n]\n")); err != nil {
```
AP-Hunt, line 102: What's this magic byte we're writing?

AP-Hunt (approved):

apiserver/server_billable_handler.go
risicle: It's the closing `]` of the json array - the top level array doesn't use `encoding/json` because that isn't compatible with streaming responses.

Streaming responses are a bit dubious in general IMHO.

risicle:
Pending approval on #3232 now

PR #184, opened 2023-07-17
added a check and unit test for jwt expiry

jackjoy-gds:
What
----

Added code to uaa_authorizer.go to throw an error if the jwt token has expired
Added a unit test for the same

How to review
-----

Look at the code
Run the unit tests
Deploy the new version and try to call the API with a known expired token 

Who can review
-----

Any paas-dev

---


AP-Hunt (approved):
I've seen that we do proper authentication of the tokens, and I'm happy that this is the most minimal change we could make

